PR #13, opened 2018-06-03
DO NOT MERGE

nemchik:


nemchik:
I'm going to keep this open and push more to it when Travis is fixed so I can confirm a few things. Everything in this branch is intentionally bad, so even when Travis is working we won't merge this.

nemchik:
I got what I wanted out of this PR, but I'm keeping it open while I communicate with Travis Support about why 4 checks happen. I would think there should only be 2 showing up. There are 2 builds happening, but each build is showing as a check twice (once with a GitHub link that links to Travis, and once with a direct link to Travis).

PR #31, opened 2018-06-12
DO NOT MERGE

nemchik:
This is just to see if travis passes.


nemchik:
This plays nicely with both branches. I'll PR this to master so that both branches can be merged independently.

PR #37, opened 2018-06-15
Move badges around

nemchik:


nemchik:
It doesn't look great on mobile. Not terrible, but not really any better than it was. I won't be heartbroken if we don't do this.

TommyE123:
Seems to display ok on mine?!

nemchik:
I'm over it. I viewed it a few more times and I've got more ideas to improve the readme, but this is fine for now.

PR #42, opened 2018-06-18
Runner

nemchik:
Trying to follow the information from here http://www.kfirlavi.com/blog/2012/11/14/defensive-bash-programming/

I still haven't actually put the main scripts into functions (the ones in root) because they are meant to be replaced by GUI menus.


nemchik:
The commit history here keeps getting messed up because of rebasing so I'm closing this in favor of #43

PR #43, opened 2018-06-18
Split installer into runners

nemchik:
Trying to follow the information from here http://www.kfirlavi.com/blog/2012/11/14/defensive-bash-programming/

I still haven't actually put the main scripts into functions (the ones in root) because they are meant to be replaced by GUI menus.

PR #45, opened 2018-06-19
Run script

nemchik:
This PR replaces #43


nemchik:
There's a lot of things going on here and I did a poor job of splitting each goal into a single commit. Here is a general overview:

- Simplify architecture detection 
- Rename yml files to accomodate simplified architecture detection
- Separate various pieces of the scripts into functions and each function into its own file
- Keep two placeholder scripts for now that act as the installer and generator (which is what the repo has always had, and will continue to have until the GUI is ready)
- Use a better method of defining `SCRIPTPATH` that works when calling the scripts from other directories
- Make certain variables like `SCRIPTPATH` read only so that they are not accidentally rewritten later
- Make certain variables local to the function they are running in so that they do not interfere with other functions
- Use a "runner" function to load the separated script files and call their functions (this is basically just a shortcut for loading and running and is not required for all scripts)
- Adjust the handling of `APPNETMODE` and `ENV_VARS` in general by the generator and new `get_env` function
- Replace `SCRIPTPATH` with `TESTPATH` in all test scripts to prevent collision with `SCRIPTPATH` in other scripts. `TESTPATH` does not use the new method to detect path like `SCRIPTPATH` does.
- Properly use `source` when a script is meant to be included and `bash` when a script is meant to be executed (this was my fault for mixing them up in some previous commits)

nemchik:
Just some other thoughts that are not part of this PR:

- In the future the two root scripts (installer and generator) would be replaced with something like `setup.sh` which should allow running with command arguments like `setup.sh -g` for the generator or `setup.sh -i` for the installer, but when run with no arguments it would present the GUI
- We could add another function in `common.sh` just like `run_script` but it would be `run_test` and have Travis call `setup.sh -t` to run the tests (which would also mean we could run them locally before pushing to GitHub and waiting for Travis to confirm things).

PR #55, opened 2018-06-22
Minor shell check Fixes

TommyE123:
Won't be a problem until Shellcheck gets updated but might as well get it covered off now.


nemchik:
For these issues https://github.com/koalaman/shellcheck/wiki/SC1117 we actually don't want literal `\` but rather we want single quoted expressions instead of double.

PR #70, opened 2018-06-26
Remove watchtower email settings

nemchik:
Watchtower is broken if these are left blank. Better to just not include them than leave it broken.


nemchik:
True. Can do.

PR #83, opened 2018-07-02
Remove version from all app yml files

nemchik:
Kind of a follow up to https://github.com/GhostWriters/DockSTARTer/commit/83c82b42b4d99c97684ec0b196f7637f7ad782fa where we found Travis was unhappy with version 3.6. The versions are not required to exist anywhere except the req files or in whatever placeholder files we have for apps without ports and things like that (although I'd like to find a better solution for those).

### DIFF
--- a/compose/.apps/bazarr/bazarr.yml
+++ b/compose/.apps/bazarr/bazarr.yml
@@ -1,4 +1,3 @@
-version:             "3.6"
 services:
   bazarr:
     image:           morpheus65535/bazarr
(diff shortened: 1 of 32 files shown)


nemchik:
Correct. We have a version number merging in by using the `.req` files, so it doesn't need to be in every other yml file we have. The first one merged takes priority. 

The whole reason I setup the `.req` files is they are the bare minimum that had to be in `docker-compose.yml` in order for `docker-compose up -d` to run (basically if you disable all apps).

PR #89, opened 2018-07-05
Apply another fix for installing compose on arm

nemchik:
https://github.com/docker/compose/issues/4874

Two things here
1. `docker-py` is apparently an issue (so, remove it!)
2. the `-U` flag should tell this section to update compose if it's already installed.


nemchik:
Had to merge master on this so it needs a second review.

Issue #71, opened 2018-06-27
VPN


ftmflea:
#### DelugeVPN
- [x] Test DelugeVPN
- [x] Make any changes required to get it DelugeVPN working
- [x] Figure out what's resetting settings on recreation
- [x] DelugeVPN fully ready to ship (check above boxes above first)
#### rTorrentVPN
- [x] Start work on rTorrentVPN
- [ ] Test rTorrentVPN
- [ ] Make any changes required to get it rTorrentVPN working
- [ ] rTorrentVPN fully ready to ship (check above boxes above first)
#### SABNZBdVPN
- [x] Test SABNZBdVPN
- [x] Make any changes required to get it SABNZBdVPN working
- [ ] Double check SABNZBdVPN running through VPN
- [ ] SABNZBdVPN fully ready to ship (check above boxes above first)
#### TransmissionVPN
- [x] Test TransmissionVPN
- [x] Make any changes required to get it TransmissionVPN working
- [x] TransmissionVPN fully ready to ship (check above boxes above first)

***

Current notes:

**DelugeVPN:**
I think my install got corrupted somehow.  I reinstalled it and it's working now.

**rTorrentVPN:**
Unable to get running so far.

**SABNZBdVPN:**
Runs. Need to check to see if running through VPN.  Ideas?

**TransmissionVPN:**
This one works.  No problems.

nemchik:
If you leave `OPENVPN_CONFIG` blank does it work?

P.S. I'm about to push a new commit to the branch so if you wouldn't mind updating your local copy (`git pull` should work) and testing again with the new changes that would be awesome!

ftmflea:
Yeah, so transmission-vpn will use a default config with PIA (not sure about others) if it can't find the config file.  I think you would need to point to it in the volumes.

ftmflea:
This works, just need to change: image: haugene/transmission-openvpn
 ```
 transmissionvpn:
    container_name: transmissionvpn
    image: haugene/transmission-openvpn
    cap_add:
    - NET_ADMIN
    devices:
    - /dev/net/tun
    restart: always
    ports:
    - ${TRANSMISSIONVPN_PORT_9091}:9091
    dns:
    - 1.1.1.1
    - 1.0.0.1
    volumes:
    - /etc/localtime:/etc/localtime:ro
    - ${DOCKERCONFDIR}/transmissionvpn:/config
    - ${DOWNLOADSDIR}:/downloads
    environment:
[...]

PR #91, opened 2018-07-09
Delugevpn

nemchik:
Per #71 DelugeVPN is ready to merge. I will prepare the other 3 in separate branches as well so we can move forward a little faster instead of waiting for all to be ready at once.


nemchik:
Yes they do, not sure how I missed those two. I'll get them added.

nemchik:
Fixed in all branches (so #92 should be in the clear now as well). Once travis passes these should be ready for another review to merge.

PR #98, opened 2018-07-16
Adjust yes/no prompts to prepare for menus

nemchik:
To trigger any of these to use the GUI we change
```
run_script 'request_reboot'
```
to
```
run_script 'request_reboot' menu
```


nemchik:
How did you get around `set +e`?

And yes the loop is needed because if someone uses the terminal prompt for `[Yn]` and they enter `z` it tells them to please enter yes or no and prompts again until they break the loop with a valid entry.

PR #103, opened 2018-07-17
Fix env_set to handle escaping

nemchik:
More GUI prep, to make sure the GUI branch can rebase/squash easily.

Previous behaviour was unhappy handling timezones because of the `/` character. This resolves by escaping.


nemchik:
Agreed about unit tests

PR #119, opened 2018-08-06
UI and Menu cleanup and input improvements

### LINKED ISSUE #118: Input improvements
Would like input menus for ID, TZ, folders, and possible other things to present a menu similar to this:
- Keep Current: `~/my_custom_docker/config`
- Use Default: `~/.docker/config` (for certain variables this should be replaced by `Use System` such as with TZ)
- Enter New

The last option should fire off a new menu that would be something like the current input prompt, but just prefill the current value (don't bother showing current/default because it would be in the previous screen) and allow the user to change the input.

https://gist.github.com/nemchik/2ba3e7514143e9260ff12096c960ffca
alt text

nemchik:
This resolves #118

This also cleans up the menu folder and consolidates most of the menu scripts into a single common UI prompting script. One of the main reasons I've done this is to eliminate nested functions and using global variables which was previously present in the menu app helper.


nemchik:
PS this has been thoroughly tested!

nemchik:
Rebasing borked the travis testing. I am going to close this and make a new PR.

PR #124, opened 2018-08-08
Install dependencies for validations on each run

### LINKED ISSUE #125: Self contained tests
Get the tests to install and update their own dependencies each run. Then remove the before section of Travis (since the validations and other scripts should be running those actions)

nemchik:
This should ensure that we're running the same version of the software (ex: shellcheck) when testing on travis or locally. it should also ensure that we're getting the latest release every time we test.

This resolves #125


nemchik:
If this is approved I will merge with squash to clean up the commit history.

PR #146, opened 2018-08-16
Remove redundant aarch64 yml

nemchik:
## Purpose
Emby, Portainer, and Watchtower do not have aarch64 images. These files should be removed since they are redundant to the armhf files which are used automatically on aarch64 systems when needed.

## Approach
Remove the files that are not needed because the generator pulls armhf images for aarch64 systems if there is no aarch64 yml file.

#### Standards
- [x] This PR meets all of the repo [standards](https://github.com/GhostWriters/DockSTARTer/wiki/Standards).


nemchik:
Tagging this as back end enhancement for lack of better label/project.

PR #167, opened 2018-08-24
Rtorrentvpn

### LINKED ISSUE #101: rTorrentVPN
See #71 for previous discussion
- [x] Add rTorrentVPN and config
- [x] Test rTorrentVPN
- [x] Make any changes required to get it rTorrentVPN working
- [x] Verify rTorrentVPN runs through VPN
- [x] rTorrentVPN fully ready to ship (check above boxes above first)

nemchik:
## Purpose
Add binhex rTorrentVPN container. This closes #101 

## Approach
Add the required files to include the app in DS. User rg9400 on Discord tested and confirmed the container to work.

- [x] Add the description for the UI and remove the version from the yml before merge
- [ ] Merge with squash
- [ ] Add to available applications after merge

## Requirements


nemchik:
I've added the commit to get the UI description and remove the version from the yml. I noticed other apps also had the version so I removed them as well. I also noticed some of the app descriptions in the UI were a little off so I tweaked those just to knock it out.

rg9400:
This works. I checked that VPN is properly loaded, but was unable to check that IP did not leak when VPN connection died. Both Flood and ruTorrent Web UIs worked. 

Few important points:
1. On Debian, you have to run `nano /etc/modules` and add `iptable_mangle` or else this will give you massive headaches like it did to me.
2. You may need to change the `LAN_NETWORK` in `VPN Settings` in `.env` to the proper LAN IP Config for it to load properly
3. You need to run the container once with Enable VPN to create the openvpn folder in the config, then move your .ovpn into that folder and recreate the container. 

I think all 3 of these can be smoothened out for users in the future. Ideally, point 1 is part of initial setup or part of the backend code for DS. Point 2 should be in App Config in my opinion, since lots of people may have differing LAN IP ranges. For point 3, a user can point to a single .ovpn file in App Config, and DS can move it into the proper config locations automatically. These are just thoughts and suggestions, but all 3 of these points created hurdles for me and were required before the container worked properly.

nemchik:
I like the idea of handling number 1 automatically in the future. I think number 2 will actually be addressed by something I already have in concept. I'm not sure about number 3, but I might make another directory in env similar to the media directories and mount it in the vpn containers automatically in the right place.

PR #171, opened 2018-08-24
Non apt

nemchik:
## Purpose
Prepare for supporting Fedora and CentOS since they are supported by the official get.docker.com script. For now they will exit the script with a fatal until we can test further, but there are other changes in this branch such as the arch detection switch that are ready to merge.

A fix for installing compose on aarch64 slipped into this branch as well.

## Approach
- Change arch detection to use `uname -m` because dpkg is only available on apt systems
- Rearrange the order of setting some things up in main.sh
- Create new update_system.sh to handle allowing updates to run through their correct package manager

This is ready to merge as is and will not enable support for Fedora or CentOS as is. We will need to remove the `fatal` lines in update_system.sh in order to open that up. Until that happens this merge will make it easy for us to either test ourselves (by removing the fatal and testing) or request testing from users willing to assist.

## Requirements


nemchik:
ping DarkSlave on discord when this merges.

nemchik:
Report from Reddit user who tested Fedora by removing the fatal that blocked Fedora from continuing: the dnf stuff in this works, but the compose install did not work. The user was able to resolve using `dnf install docker-compose`. I have asked for a little more information from the user and will add the required fixes in a later PR.

nemchik:
For reference this build is fine but we should probably merge #176 first since Codacy is now offering the full array of shellcheck 0.5.0 validations.

nemchik:
This whole thing will be merged with squash at this point since the commits have become a little messy.

PR #173, opened 2018-08-27
Travis failures

nemchik:
## Purpose
Use `allow_failures` in `.travis.yml` on the `run_generate` step so that `fast_finish` actually has an effect.

## Approach
We still want to test the generator and fail a build if it doesn't work, so I have split `run_generate` and created `run_generate_full` which is identical to what `run_generate` used to be and I have slimmed `run_generate` down to just portainer and watchtower. Builds complete much faster now (under 3 minutes as opposed to around 10) and we can still review the builds to see if `run_generate_full` fails, which would be caused by an individual app having trouble.

- [ ] Merge this with squash.

#### Learning
https://docs.travis-ci.com/user/customizing-the-build/#rows-that-are-allowed-to-fail
https://docs.travis-ci.com/user/build-stages

## Requirements

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,8 +13,10 @@ jobs:
     script: sudo bash ./main.sh -t run_install
   - stage: run
     script: sudo bash ./main.sh -t run_generate
+  - stage: run
+    script: sudo bash ./main.sh -t run_generate_full
 matrix:
   fast_finish: true
   allow_failures:
   - stage: run
-    script: sudo bash ./main.sh -t run_generate
+    script: sudo bash ./main.sh -t run_generate_full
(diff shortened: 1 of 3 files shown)


nemchik:
I'm open to discussing this change as I can see there being an argument against this method where we would not always be fully testing a new app added to the repo to ensure the container fires up properly. The issue could be with the container itself, or we could have made an error in the yml we include in DS that should not be merged to master. We would have to be more diligent to check the full build logs for new apps. The positive trade off is for changes that generally only affect the bash scripts builds will finish in 1/3 the time while still running the full test in the background.

Thoughts?

PR #210, opened 2018-09-06
Remove orphans with compose up

nemchik:
## Purpose
Revisit https://feathub.com/GhostWriters/DockSTARTer/+14 and actually remove orphans using the build in option with compose. The prune option that was considered to resolve the request does not actually remove orphans, it only cleans up unused objects after the container is stopped. Without using the build in remove orphans option the container is left running.

## Approach
Add the remove orphans option to compose up. This will Stop and remove a container that is no longer considered part of the stack. The existing prune option will clean up unused resources such as the image that the container used (this is not automatically removed when an orphan container is removed).

#### Learning
https://docs.docker.com/compose/reference/up/

## Requirements


nemchik:
Pinging @doowrednu as this request originally came from him on feathub and he may want to follow this PR for updates when it merges.

Issue #186, opened 2018-09-02
Individual App Settings


nemchik:
![preview](https://media.discordapp.net/attachments/477959324183035940/483419446334062607/Screenshot_20180809-193614.png)

`whiptail --yesno "$(grep 'PLEX_' ~/.docker/compose/.env)" 0 0`

```
while IFS= read -r line; do
    SET_VAR=${line/=*/}
    SET_VAL=${line/*=/}
    echo
    echo "SET_VAR = ${SET_VAR}"
    echo "SET_VAL = ${SET_VAL}"
    echo
done < <(grep "${APPNAME}_" ~/.docker/compose/.env)
```

PR #238, opened 2018-09-14
Add airdcpp

### LINKED ISSUE #234: Airdc++
Airdc++: 

Info about it: 

https://www.airdcpp.net/

https://en.wikipedia.org/wiki/DC%2B%2B

Docker and integration:

The best existing docker could be https://hub.docker.com/r/romancin/airdcpp-webclient/

As we can see they only need one port, two bind for persistent data (one for download and another for sharing content). To work, need 2 files before starting the docker, those two files are accesible from docker github.

Reason 

its a DC++ client, a new way to get new content for the media server.

JuezFenix:
## Purpose
Add Airdc++ Web Client docker. Close #234

- [ ] Verify config files are created when the container is run
- [ ] Add to available apps wiki after merging

#### Learning

https://hub.docker.com/r/romancin/airdcpp-webclient/

## Requirements


compose/.apps/airdcpp/airdcpp.yml
```diff
@@ -0,0 +1,15 @@
...
+    restart:         always
+    environment:
+      - PGID=${PGID}
+      - PUID=${PUID}
+      - TZ=${TZ}
+    volumes:
+      - /etc/localtime:/etc/localtime:ro
+      - ${DOCKERCONFDIR}/airdcpp:/.airdcpp
+      - ${DOWNLOADSDIR}:/Downloads
+      - ${AIRDCPPSHAREDIR}:/Share
```
nemchik: Looks like this should be `AIRDCPP_SHAREDIR`

scripts/generate_yml.sh
```diff
@@ -70,6 +70,11 @@ generate_yml() {
             fi
         else
             error "Could not find ${SCRIPTPATH}/compose/.apps/${FILENAME}/ directory."
+        fi
+		if [[ ${FILENAME} == "AIRDCPP" ]]; then
+			echo "Downloading needed config files for Airdd++ Client"
+			wget https://raw.githubusercontent.com/romancin/airdcpp-webclient-docker/master/.airdcpp/DCPlusPlus.xml -O ${DOCKERCONFDIR}/${FILENAME}/DCPlusPlus.xml
+			wget https://raw.githubusercontent.com/romancin/airdcpp-webclient-docker/master/.airdcpp/WebServer.xml -O ${DOCKERCONFDIR}/${FILENAME}/WebServer.xml
```
nemchik: I'm not a big fan of the wget lines being added to the generator. Does the container not put these in place the way it should? Maybe that should be a bug opened with the container dev?

scripts/generate_yml.sh
```diff
@@ -70,6 +70,11 @@ generate_yml() {
             fi
         else
             error "Could not find ${SCRIPTPATH}/compose/.apps/${FILENAME}/ directory."
+        fi
+		if [[ ${FILENAME} == "AIRDCPP" ]]; then
+			echo "Downloading needed config files for Airdd++ Client"
```
nemchik: Spelling

scripts/menu_app_select.sh
```diff
@@ -40,6 +40,8 @@ menu_app_select() {
                 APPONOFF="off"
             fi
             case "${APPNAME}" in
+                "AIRDCPP")
+                    APPLIST+=("Airdc++" "Web Client to connect to DC++ Hubs" "${APPONOFF}")
```
nemchik: This line will have to read:
```
APPLIST+=("AirDCpp" "Web Client to connect to DC++ Hubs" "${APPONOFF}")
```
and cannot use the `++` due to the way naming works in DS. Caps don't matter, so that could be changed to something like `airdcPP` or whatever makes the most sense.

scripts/menu_value_prompt.sh
```diff
@@ -30,6 +30,10 @@ menu_value_prompt() {
             ;;
         DOWNLOADSDIR)
             SYSTEM_VAL="${DETECTED_HOMEDIR}/Downloads"
+            VALUEOPTIONS+=("Use System" "${SYSTEM_VAL}")
+			;;
+        AIRDCPP_SHAREDIR)
+            SYSTEM_VAL="${DETECTED_HOMEDIR}/shared"
```
nemchik: This will actually tie into #186 which I am planning to tackle next month. For now let's leave this out and users will be required to edit their `~/.docker/compose/.env` file to set this directory.

nemchik:
Side note, we'll need to get this to pass Travis-CI as well. It seems like it's complaining about newlines at the end of the YML files.

scripts/generate_yml.sh
JuezFenix: Not yet, we can wget them or copy from local to the config dir

compose/.apps/airdcpp/airdcpp.yml
JuezFenix: Yes, my mistake.

scripts/menu_app_select.sh
JuezFenix: Ok, and i see i missed ;; to close the case :)

scripts/menu_value_prompt.sh
JuezFenix: Airdc++ use two folder one for downloads and other for shared content. Dowloads shared automaticaly, but other content only can be shared using this folder.

nemchik: I'm fine with the folder being a part of `.env` and the yml for the app, but I already have a plan to prompt for app settings in another way. The changes specifically to `scripts/menu_value_prompt.sh` won't be needed once #186 is developed (planned for next month).

JuezFenix: So, for now still this way rigth?

nemchik: While this method is not incorrect, it would be removed later. Everything else in this file will remain even after #186 merges. I would prefer the changes to this file be reverted. In the meantime, users will be able to edit their `.env` file to set this the way they would like.

PR #272, opened 2018-09-27
app: traefik

### LINKED ISSUE #236: traefik
traefik: 

Info about it: 

https://traefik.io/

Docker and integration:

Traefik have a official docker container, they only need one file (config), a bind to docker socket and 2 ports. Traefik is completely dynamic and only need a few tags on the containers to know all the details about wath to do (example: port, protocol and subfolder).

Reason 

Can be used to transforms a tons of ports to a hierarchy of subdomains or folders. As well can be used to use the same credentials to all services and protect that not support it (like h5ai). In addition, another thing that can be done is to use the same certificate for the entire environment (requested on traefik first run using the acme integration).

robertbaker:
## Purpose
This adds traefik as an option with the minimum amount of changes + the most flexibility.
## Approach
By default:
Auto expose docker containers to traefik, auto generate hostnames. appname.compose.mydomain.tld (there's no way to remove the compose. subdomain w/o labels).

Port 80 and 443 out of the box. Recommended to disable http and enable ACME for letsencrypt.
It's also recommended by traefik to remove - --api  and 8080 in prod. The dashboard they give is minimal, I may disable it by default.

- [x] should 80/http be disabled by default?
- [x] should the dashboard be disabled by default?

#### Learning
https://docs.traefik.io/user-guide/docker-and-lets-encrypt/
https://docs.traefik.io/configuration/acme/#provider
https://github.com/joshuhn/media-docker/blob/master/docker-compose.yml

an example acme challenge
```
TRAEFIK_ACME_CHALLENGE=--acme.dnschallenge
TRAEFIK_ACME_CHALLENGE_PROVIDER= --acme.dnschallenge.provider=cloudflare
```

then you need to add these 2 env vars to `traefik.override.yml` file for cloudflare 

```
services:
  transmission:
    environments:
      - CLOUDFLARE_EMAIL=${CLOUDFLARE_EMAIL}
      - CLOUDFLARE_API_KEY=${CLOUDFLARE_API_KEY}
```

Modify .docker/compose/.env

```
add:
CLOUDFLARE_EMAIL=
CLOUDFLARE_API_KEY=
```

In a appname.override.yml

```
    labels:
      - "traefik.backend=appname" # optional
      - "traefik.frontend.rule=Host:appname.domain.org"  # optional, default is appname.compose.domain.org
      - "traefik.enable=true" #only if TRAEFIK_EXPOSED_DEFAULT is false
```

For more providers read: https://docs.traefik.io/configuration/acme/#provider

Please note: 

TLS-ALPN-01 challenge is disabled for letsencrypt due to security issues. I recommend dns challenge so you don't need port 80 open on your host.

CloudFlare does not support second-level subdomains when using cloudflaire proxy /w universal ssl  + letsencrypt, which are created by default. To get around this, you'll need to add a label to your apps to remove the compose subdomain or disbale cloudflare proxy (not recommended).

## Requirements

closes #236


robertbaker:
@nemchik ready for review

nemchik:
At first look, you've duplicated the env example into the req folder.

Otherwise just to confirm you have tested this and it works and you don't need to modify other apps yml files to get the proxy setup?

robertbaker:
@nemchik thanks, fixed.

I confirmed it's working with the default env settings. I also confirmed turning on acme, only allowing https, doing dns challenge with cloudflare. I confirmed it "auto-configures" for apps. The defaults are appname.compose.domain.tld. I couldn't figure out how to remove 'compose' from the url without using labels.

nemchik:
To address your pre-merge questions:

- I would keep port 80
- if the dashboard is exposed publicly with no way to add some kind of authentication then I'd want to discuss why anyone would need it (what does it do?)
- if you can do something like set a password to access the dashboard then I'd keep it
- if the dashboard is only accessible from the local network and not available from a public address then I'd keep it

nemchik:
As for `compose` being in the domain I'm not in love with it, but if the options are that compose in the name or labels I'll take compose in the name.

Labels in DockSTARTer would either be always included even even not needed, or it'll require some extra coding to make it only include if traefik is enabled. It also means more yml to maintain for each app. I'm not in a hurry to do either of those things.

robertbaker:
The dashboard is only on the local network.

Compose in the url here is coming from the "stack name". It can be changed,
but the only way to remove the second level subdomain is labels, in this
case, it's just 1 label for each app.

Other apps seem to be using labels for config.

Hera is a cloudflare Argo container that uses labels to assist with auto
config. I was looking into using it but Argo is a paid cloudflare service.
It requires 2 labels.

I have an idea of how to handle these situations and provide support for
custom "templates" or stacks, to provide more of a complete OoTB
experience. Use git branches that only add override files to apps. This
would be a new level of override in between orgin and user. A menu option
that traverses git branches and pulling would be all that's needed. Every
time master is updated the stack branches are just FF without conflicts.

nemchik:
I actually think this is where users should be making user of overrides on their own to pick how they want this handled. People can use `docker-compose.overridd.yml` to add all the labels they would like in one place.

nemchik:
Only other thing would be if you wouldn't mind adding a few more ports to adjust the run_generate_full test so that the `*.4` builds pass https://travis-ci.com/GhostWriters/DockSTARTer/jobs/148780604

These are set to allow failure because they take forever to complete, but I try to make sure new apps pass the full test.

nemchik:
I've been thinking a lot about this and how having Traefik in DockSTARTer will affect the project, and what support we can offer, and how this may result in other app devs having increased support requests regarding Traefik with their apps.

At the moment I don't feel like adding Traefik will be a positive experience for users of DockSTARTer. Our staff lacks the knowledge or experience to properly support it.

If there are changes made to Traefik that resolve some of the concerns that I have with it we may revisit it in the future.

PR #276, opened 2018-09-29
 expose communication ports to docker containers

robertbaker:
## Purpose
Expose allows other docker containers to  access other docker containers.
Although this isn't required when using port, if you we're using traefik,  you won't need to publish ports to the host! This may be useful for https://github.com/aschzero/hera too

## Approach
_How does this change address the problem?_
Adds expose as a way to document the primary inner communication ports.

- [ ] how to disable ports with override?
- [ ] Should these be appname.expose.yml instead of being in ports?

#### Learning
https://medium.freecodecamp.org/expose-vs-publish-docker-port-commands-explained-simply-434593dbc9a3

## Requirements

### DIFF
--- a/compose/.apps/airdcpp/airdcpp.ports.yml
+++ b/compose/.apps/airdcpp/airdcpp.ports.yml
@@ -6,3 +6,6 @@ services:
       - "${AIRDCPP_PORT_21249}:21249"
       - "${AIRDCPP_PORT_5600}:5600"
       - "${AIRDCPP_PORT_5601}:5601"
+    expose:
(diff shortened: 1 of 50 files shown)


nemchik:
The way I understand it, expose should be done in the dockerfile for each image. While it can be done in compose it seems like that's only needed as an afterthought if a container doesn't expose a port in its dockerfile and another app needs to communicate to it on that port.

I'm going to read more into this and see what would make the most sense for how to handle this.

nemchik:
Does this have no relevance without Traefik?

robertbaker:
I believe it's redundant

PR #430, opened 2018-12-26
InfluxDB and Telegraf

### LINKED ISSUE #375: Influxdb

nemchik:
## Purpose

This closes #375 
Grafana is not very useful without InfluxDB. Telegraf usually goes hand in hand with it for various things.

## Approach

Add the required files for both.

- [x] Confirm stage 4 tests pass

#### Learning

https://hub.docker.com/_/influxdb
https://hub.docker.com/_/telegraf

## Requirements


compose/.apps/telegraf/telegraf.yml
```diff
@@ -0,0 +1,16 @@
...
+    container_name: telegraf
+    restart: unless-stopped
+    environment:
+      - HOST_PROC=/host/proc
+      - TZ=${TZ}
+    volumes:
+      - /etc/localtime:/etc/localtime:ro
+      - /proc:/host/proc:ro
+      - /var/run/docker.sock:/var/run/docker.sock:ro
+      - ${DOCKERCONFDIR}/telegraf:/var/lib/telegraf
```
Antexa: - ${DOCKERCONFDIR}/telegraf:/var/lib/telegraf

Don't u think of using /etc/telegraf could be a better location ? According to the documentation of Telegraf - https://hub.docker.com/_/telegraf

nemchik: The issue with `/etc/telegraf` is there are other files aside from just the `telegraf.conf` in that folder. Those other files should be part of the container and should update when the container updates. I could have DS mount just the `/etc/telegraf/telegraf.conf` file, but if the file does not exist outside the container before attempting to start the container then docker-compose will create an empty folder named `telegraf.conf`. This is not something I can change other than trying to create an empty file ahead of time, but in most cases I don't think an empty conf file is too useful.

Also you can use environment variables to set everything that `telegraf.conf` contains, and even if you were to have the file mapped as a volume the environment variables would take precedence. You can set those variables by reviewing their docs to determine the naming and then use our https://dockstarter.com/advanced/overrides/ guide to get them in place.

nemchik: p.s. I am open to alternatives if there are documented best practices.

Antexa: A small trick : docker run --rm telegraf -sample-config > telegraf/telegraf.conf

https://www.jacobtomlinson.co.uk/monitoring/2016/06/23/running-telegraf-inside-a-container/

Issue #436, opened 2018-12-30
Update calibreweb to calibre-web


Ssnarf:
**Describe the bug**
When adding calibreweb error received:
`pull access denied for lsioarmhf/calibreweb, repository does not exist or may require 'docker login'`

Looks like the image name is incorrect in calibreweb.armhf.yml hence the pull error.

**To Reproduce**
Select App - CalibreWeb
Build containers

**System (please complete the following information):**
- OS: [e.g. Debian]
Raspbian Stretch RasPi 3B+
Linux raspi 4.14.79-v7+ #1159 SMP Sun Nov 4 17:50:20 GMT 2018 armv7l GNU/Linux

nemchik:
Sorry I didn't reply to this sooner! I wonder if there was an issue happening with docker hub at the time. https://hub.docker.com/r/lsioarmhf/calibre-web/ this seems to still exist and is not marked as deprecated. I know LSIO is moving their images to a new build system but this one doesn't seem to be done yet. Could you try again to run the container?

PR #445, opened 2019-01-05
Added hyphen

Ssnarf:
Updated docker hub repo address to solve pull error highlighted in #436.
This might also be necessary for the aarch64 file but I do not have the hardware to test.


nemchik:
Wow! Good catch! It looks like arm64 (aarch64) needs this as well, would you mind adding it to your PR?

Ssnarf:
Sure, will do it now.

nemchik:
Thanks again for this!

PR #447, opened 2019-01-06
LogitechMediaServer

karl0ss:
This is a working implementation of LogitechMediaServer (Squeezebox).

There is a note that needs to be added to the wiki, I can add this if needed...

- MUST use port 9000 as this is used for the communication of from your physical players around your house, meaning Portainer needs to be moved if you install this.

- You should change the LOGITECHMEDIASERVER_MUSICDIR var to the location of your music, and then set your location to /music within LMS

Config is stored in the standard {DOCKERCONFDIR}/logitechmediaserver


compose/.apps/logitechmediaserver/logitechmediaserver.ports.yml
```diff
@@ -0,0 +1,15 @@
+services:
+  logitechmediaserver:
+    ports:
+      - "9000:9000"
+      - "9090:9090"
+      - "3483:3483"
+      - "3483:3483/udp"
```
nemchik: Ports need to use variables like all other apps. See the Plex template for an example of how UDP is handled with variables).

compose/.apps/logitechmediaserver/logitechmediaserver.ports.yml
```diff
@@ -0,0 +1,15 @@
...
+      
+      
+      
+      
+      
+      
+      
+
```
nemchik: Only one blank like at the end of the file is needed.

compose/.apps/logitechmediaserver/logitechmediaserver.yml
```diff
@@ -0,0 +1,11 @@
+#/APPNICENAME=LogitechMediaServer
+#/APPDESCRIPTION=LogitechMediaServer(SqueezeBox)
+services:
+  logitechmediaserver:
+    container_name: logitechmediaserver
+    image: larsks/logitech-media-server
+    volumes:
+      - /etc/localtime:/etc/localtime:ro
+      - ${LOGITECHMEDIASERVER_MUSICDIR}:/music
```
nemchik: Can this use `MEDIADIR_MUSIC`?

compose/.apps/logitechmediaserver/logitechmediaserver.ports.yml
karl0ss: As far as I'm aware, Lms will not work on any other ports due to the way that the "phisycal players" communicate with the server, they are hard coded to look at these ports.

Beause of that I didn't make em configurable and would add a note to the wiki that if Lms is used then Portainer needs to move from port 9000.

compose/.apps/logitechmediaserver/logitechmediaserver.yml
karl0ss: Yep, will fix this, didn't think about that, but I guess mediadir_music is where all your music should be sorted, so that makes sense...will update when I get home

compose/.apps/logitechmediaserver/logitechmediaserver.ports.yml
nemchik: For the sake of keeping the template in line with the all the others this should still use variables, but the default port in the `.env.example` file can be `9000` and like you said a wiki entry can be created for this app to explain that it requires port `9000` and if the user is using portainer they would need to move that to another port. There are still valid reasons why the user might edit the port, so it should be an option. Also the other ports likely don't have to be hard coded and should be variables.

nemchik: Awesome! I appreciate the PRs!

PR #442, opened 2019-01-05
Jdownloader

karl0ss:
This is my implementation of jdownloader2 in our DockSTARTer,

This works on my install, and I have copied how other apps have appeared to be added to the system, but this doesn't appear to allow my app to be installed? its not present on the list...

Any help would be good please

Issue #410, opened 2018-12-04
JDownloader2 container


trohnjavolta:
**Is your feature request related to a problem? Please describe.**
No
**Describe the solution you'd like**
JDownloader is a open source download manager written in java. It's especially nice for use with one click hosters.
I'm successfully using this multiarch container maintained by jaymoulin: https://store.docker.com/community/images/jaymoulin/jdownloader

juligreen:
I like the idea, but I'm not certain that this makes sense in the context of this project.
For what are you using JDownloader?
Does it integrate into other services on this project?

trohnjavolta:
I'm using it as a downloadmanager, like nzbget or transmission, deluge etc.
How 

> Does it integrate into other services on this project?

No but neither do other apps supported in dockstarter

trohnjavolta:
> I wanted this as well, So have built it...
> 
> Will be added once this PR is merged - #448

Great but can you explain why you use jlesage/jdownloader-2 docker container and not jaymoulin/jdownloader multiarch container?
I'm fairly new to docker, jaymoulin/jdownloader works good for me on armhf.
Does jlesage/jdownloader-2 even work on arm architecture? I can't find any hints in the dockerfile that it does. + I'm confused about the used java runtime environement. It says openjre but because archive extractor doesn't work it uses oracle jre...Well, does it use both or what?
I just can speak for armhf platform, I read multiple times that oracle jre works better on it with jdownloader 2.

I'm certain you have way more knowledge of all this than I do. Would be great if you could explain your container choice.

karl0ss:
I will have a look into if this image works on arm, I will try deploy it to my raspberry pi tonight, the main reason I prefer this is it exposes the JD GUI via port 5800 so you don't have to use the daemon via cli or use the webui all the time (though you can).

Archive extracting for me works in this container..

Let me see if it will deploy to arm later and I will update this post..

trohnjavolta:
> I will have a look into if this image works on arm, I will try deploy it to my raspberry pi tonight, the main reason I prefer this is it exposes the JD GUI via port 5800 so you don't have to use the daemon via cli or use the webui all the time (though you can).
> 
> Archive extracting for me works in this container..
> 
> Let me see if it will deploy to arm later and I will update this post..

Ok, so there's a "full blown gui" running in the container? I'd prefer a headless instance since I guess it uses up less ressources and the myjd webui works great. But yeah, that's just my 2 cents.
[...]

PR #448, opened 2019-01-07
Jdownloader2

karl0ss:
Working PR to enabled JDownloader2 in DOCKstart

- Port 5800 is used to access the JDownloader GUI via http://<SERVERIP>:5800
- Port 3129 is used for the myJdownloader client to connect.

- /download is using the ${DOWNLOADSDIR}, this needs to be set in jdownloader settings once 
container is running via http://<SERVERIP>:5800, I can add a wiki entry for this
- /config is pointing to the standard ${DOCKERCONFDIR}/jdownloader2


nemchik:
This looks good. I'm going to retry the travis builds to see if I can get them to pass (travis is weird about passing PRs because of github API limits)

karl0ss:
No problem, I just pulled my branch with this and my LogitechMediaServer PR and both are showing and installing as expected..

nemchik:
Can we keep this called `jdownloader2` rather than `jdownloader2_gui` for simplicity? (does that still accurately represent the name of the app?)

nemchik:
Reference #442 #410

karl0ss:
I have updated the app as I have now also created the multiarch version that was spoken about in the app request, and called that _nogui, there is a pr for both.

nemchik:
I feel like having two separate jdownloader2 images is a bit confusing and I think it would be best if we could find one image to use that offers the most to us.

My (loosely defined) preferred ranking of images would be:
1. Official image from the dev of the app contained inside the image
2. Image published by a well known reputable source like LSIO/binhex/oznu/hotio
3. The most popular available image on docker hub that offers support for multiple architectures and is up to date (updated at least as often as the app inside the container)
4. The easiest to use image available from docker hub that otherwise meets all the criteria of the point above with the exception that it might not be the most popular but instead offers legitimate and valid ease of use improvements or is better maintained (more recently updated or kept up to date with the app contained in the image)

I'll see if we can locate such an image and post it in the relevant issues/PRs and maybe we can just stick to one. If I'm missing something and there's a valid reason for these to be two separate images (ex: deluge without a vpn and deluge with a vpn) please let me know.

nemchik:
I'm sure something like this would be possible. I'll ask some friends what their thoughts on it are, and will see if we can locate an already existing candidate that meets all the needs.

PR #451, opened 2019-01-07
Kodi headless

karl0ss:
Very useful if you use kodi and synchronized library.


nemchik:
Can you add the ARMHF and ARM64 images? Also if you notice the minor tweaks to the other PRs where I adjusted the ports and reordered (sorted) some of the items. If you could do that stuff we'd be in good shape for this PR.

https://hub.docker.com/r/lsioarmhf/kodi-headless-aarch64
https://hub.docker.com/r/lsioarmhf/kodi-headless

karl0ss:
added the other versions.

No this one doesn't, it isnt a true install of kodi, its a headless build that is usefull for using as you "host" client to recieve update commands from medusa and such, normally if you are using a mysql server to host your database for multiple clients.

This is all done in kodi via the advancedsettings file, using kodi's inbuilt methods fo the file source, the kodi config dir, i have set into the ds config dir..

karl0ss:
I will make the changes you have asked for in the commits, I am currently building my own version of this docker container, as it is missing some useful stuff, so once I have done that, I will update this PR with that and the fixes.

nemchik:
I've made some minor tweaks to update this branch with new changes from master. There's still the other minor things I commented on requesting changes. Once those are complete I'll begin working to get a passing build so this can merge.

nemchik:
The tweaks I mentioned above are still needed, but now I believe lsio has switched the image too use their new pipeline build system which means for ARM images we need the same image as x86. Also I've separated the x86 image into it's own file in all other apps, so it'll need to be done here too.

nemchik:
After further investigation it seems like lsio have not moved their Kodi image to their new build system, and this PR hasn't had the requested changes made so I'm going to close it. Feel free to open a new PR for this to be added if the changes are made (reference other PRs that add apps).

PR #493, opened 2019-01-21
Compose fixes

nemchik:
## Purpose

Use a more universal method of installing `docker-compose`
This closes #490

## Approach

* Make efforts to remove former `docker-compose` installations
* Install `python` and `pip` (and dependencies as needed)
* Use `pip` to install `docker-compose`

#### Open Questions and Pre-Merge TODOs

- [x] Rebase/squash
- [x] Confirm someone on ARM can run the testing commands below
- [ ] Confirm someone on CentOS can run the testing commands below
- [x] Confirm someone on Fedora can run the testing commands below

To test upgrading an existing installation:
```bash
cd ~/.docker
git fetch -p
git reset --hard origin/compose-fixes
git pull
sudo ds -e
# Enter password when prompted
sudo ds -i
# No need to reboot
sudo ds -c up
```

To test a clean installation:
```bash
sudo apt-get install curl git
git clone https://github.com/GhostWriters/DockSTARTer ~/.docker
cd ~/.docker
git fetch -p
git reset --hard origin/compose-fixes
git pull
sudo bash main.sh -i
```

#### Learning

https://docs.docker.com/compose/install/#install-using-pip
https://cryptography.io/en/latest/installation/#building-cryptography-on-linux
https://urllib3.readthedocs.io/en/latest/advanced-usage.html#ssl-warnings
https://urllib3.readthedocs.io/en/latest/user-guide.html#ssl
https://urllib3.readthedocs.io/en/latest/user-guide.html#ssl-py2

## Requirements

- [x] These changes meet the standards for [contributing](https://github.com/GhostWriters/DockSTARTer/blob/master/.github/CONTRIBUTING.md).
- [x] I have read the [code of conduct](https://github.com/GhostWriters/DockSTARTer/blob/master/.github/CODE_OF_CONDUCT.md).

PR #507, opened 2019-01-26
Adjust pip failures

nemchik:
## Purpose

Fix compose install failure due to pip issues.

## Approach

The pip dependencies will try to install but are ok to fail because compose itself will also try to install its own dependencies. DS installing pip dependencies is more of a helping hand than a requirement.

## Requirements


nemchik:
This was a follow up to #493

PR #570, opened 2019-02-17
Add host GPU into the Plex container for HW accelerated decoding

ohthehugemanatee:
If the host machine has hardware acceleration available, this will make it available in the container. If not, /dev/dri should be a null device and get ignored. Probably also depends on the user running the container being in the "video" group.

## Purpose

Plex (with a plex pass) supports hardware transcoding. But your host machine's GPU device has to be accessible from inside the container! This reduces CPU usage for transcoding _significantly_.

## Approach

Maps the `/dev/dri` device into the container. This will make any GPU with open source drivers (read: Intel, AMD, but not NVIDIA) available inside the container.

- [ ] Doing this for NVIDIA is a lot harder. You set it up on the host and they provide an "NVIDIA drivers" volume which you mount into the container... Need to decide if we bother or not.
- [ ] Need to update the wiki for Plex to say the UID running the container needs to be in the `video` Group, so it has permission.
- [ ] I understand that recent versions of ffmpeg use hardware acceleration if it's available... so maybe this is useful even for people who don't have a plex pass. :)
- [ ] Is this helpful for Emby, too?

## Requirements


nemchik:
Is there any documentation on the official Plex image or lsio to support this as the best way to do it?

Does the video group always exist? If so we might like DS to just take care of adding your user to that group.

As long as we're sure that users without hardware acceleration won't be affected by this change that's good.

I think the formatting of the yml is incorrect. The line starting with the `-` needs to be indented.

ohthehugemanatee:
* Fixed indentation to match the rest of the file
* The group which controls access to /dev/dri is named differently on different distributions. On Ubuntu it's `video` but I don't know about others. We could script detecting the group owner of `/dev/dri`, adding the user to that group, and validating group permissions. 
* I can't find any specific documentation, but I wouldn't expect there to be any. `/dev/dri` is where GPU communication lives in Linux, period. On some further research, this isn't just the GPU from the ATI/Intel drivers; it's the entire Direct Rendering Manager, the Linux kernel component which controls GPU communication. It doesn't vary across distributions or hardware configurations. NVidia cards have to use it just like everyone else. The Linux kernel includes free/open source support for all three major manufacturers, and a long list of alternatives. See [Wikipedia](https://en.wikipedia.org/wiki/Direct_Rendering_Manager#Hardware_support) for the complete list. 
* `/dev/dri` should always exist on Linux devices, and offers whatever GPU resources are available. So there's no special case required for devices that don't have hardware decoding; Plex handles that itself.

nemchik:
Ok, so this should be pretty much ready to merge if we include info in the wiki about how to find the correct owner group and add your user to that group.

The Travis tests are picky on PRs but I'll merge master and rerun until they pass.

nemchik:
Got lucky with the build passing. Would you agree about my comment above about the wiki?

Issue #383, opened 2018-11-15
Verbose output


nemchik:
Make something actually handle the verbose flag. A helper function that only redirects output (to make it silent) if a verbose flag is set would be ideal. Also, use `tee` to log the output if it's not being silenced

Maybe make verbose levels along the lines of
- 0 is silent
- 1 basic logging only to the log file
- 2 basic logging to the log file and console
- 3 verbose logging to the log file, basic logging to the console
- 4 verbose logging to both the log file and the console

Verbose logging would redirect to `tee` or to the log file

nemchik:
https://stackoverflow.com/questions/2520085/how-do-i-conditionally-redirect-the-output-of-a-command-to-dev-null

nemchik:
Keep this

PR #577, opened 2019-02-21
Switch to universal prompt code

### LINKED ISSUE #382: Silent/quiet/yes arg
Pick one and make it a thing. No auto reboot. Travis can use this flag rather than needing special code all over the place to skip certain things when running as Travis (basically replace all the Travis considerations with silent/quiet/yes)

Consider that the default desired response might not always be yes. Maybe wording of the prompts needs to change so that the default yes response makes the most sense.

Silent/quiet maybe should not be used because they imply no output (meaning no logging output to stdout, just send it to the file). This might be better handled by the verbose flag implementation.

nemchik:
## Purpose

This closes #382 
Default to not prompt for most things
Use sane default for opting into prompts as needed
Remove the actual reboot code and instruct the user to manually reboot

## Approach

Create a standard question prompt function since we're reusing that logic a lot. Allow the question prompt function to prompt as CLI or GUI (although we currently do not have any need for CLI prompts, keeping the option for the future). Get rid of the code that performs a reboot and replace it with information requesting that the user perform the reboot manually along with a warning stating the importance of a reboot on first run.

Edit: I've also added a significant rework of any place we used a `yesno` with `whiptail` so that it all uses the same standard code. There's also some minor renaming to ensure that naming conventions are proper (slipped by me in the past). Also Travis runs fine with the built in compose code so we no longer run compose commands separately in the test script (we just allow the test script to run compose through ds).

In the future we may want to look into getting rid of `INTERFACE` and `PROMPT` code and just define it once for the whole script rather than individually for each function. We had done this in the past but undone it with this PR for now just to get this moving. There's no functional reason to have to get rid of it, just a personal preference on my part.

- [x] Testing

## Requirements

### DIFF
--- a/.scripts/cmdline.sh
+++ b/.scripts/cmdline.sh
@@ -54,22 +54,22 @@ cmdline() {
             c)
                 case ${OPTARG} in
                     down)
-                        run_script 'run_compose' down
+                        run_script 'run_compose' "${INTERFACE:-false}" down
                         ;;
                     generate)
                         run_script 'generate_yml'
                         ;;
                     pull)
                         run_script 'generate_yml'
-                        run_script 'run_compose' pull
+                        run_script 'run_compose' "${INTERFACE:-false}" pull
                         ;;
                     restart)
                         run_script 'generate_yml'
-                        run_script 'run_compose' restart
+                        run_script 'run_compose' "${INTERFACE:-false}" restart
                         ;;
                     up)
                         run_script 'generate_yml'
-                        run_script 'run_compose' up
+                        run_script 'run_compose' "${INTERFACE:-false}" up
                         ;;
                     *)
                         fatal "Invalid compose option."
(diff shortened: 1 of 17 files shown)


nemchik:
Reference #382 to be closed by this PR since the issue would no longer be relevant after these changes. The idea of silent would be addressed with #383 instead.

mattgphoto:
Based on my own testing of these steps:
```
1. ds -c
2. ds -c up (or down, or any other thing like that)
3. ds -u (should set you back to master without prompting)
4. ds -u origin/prompt (should set you back to the prompt test without prompting)
5. ds -p
6. ds -i (should not prompt at all, but should display a message asking first time users to reboot 1.no actual reboot is ever run)
7. using the DS menu to update DS (should set you back to master with a GUI prompt)
8. using the DS menu to go through the config menu and get to a point where the GUI prompts you to run compose (app select with no changes is probably shortest)
9. using the DS menu to install dependencies (should end with the CLI messages asking the user to reboot if it's their first run)
10. anything else you can think of where DS used to prompt you to do something either via CLI or in the GUI with a yes/no
```
All test tasks PASSED.

Issue #524, opened 2019-01-31
Test initial install


nemchik:
Test what the readme shows as the first time install.

PR #587, opened 2019-03-01
Inline testing

### LINKED ISSUE #387: Test backups

### LINKED ISSUE #523: Test updates

nemchik:
## Purpose

This closes #523 
This closes #387 

Each function file should be tested individually wherever possible to ensure they are able to perform their intended function. Exceptions to this would be things Travis cannot test such as GUI. This is a solid step in the right direction, but there is still more that could be done. Right now the logging and lack of failure is considered proof of success. In the future we may implement additional checks to confirm success.

## Approach

Create a `test_FUNC` function in each file in `.scripts` (ex: `test_backup_create`). Adjust the `run_test` function in `main.sh` to run tests from the files in `.scripts` rather than separate tests in `.tests`. Adjust Travis to look for each file in `.scripts` and run that file's `test_FUNC`. Travis will fail if a `test_FUNC` is missing. Files who have nothing Travis is able to test still have `test_FUNC` functions but they are dummy functions that simply display a warning and do not cause builds to fail.

## Requirements


nemchik:
Reference #524 
This PR does NOT close it because it asks for something different (the readme install method) to be tested. This PR makes testing the readme install method require a different approach (cannot be done in a `.sh` file in the repo but would need to be done only in `.travis.yml`).

Issue #599, opened 2019-03-14
bug: symlink for ds wrong when path does not start with /


phenonymous:
### The issue:
If you invoke main.sh by relative path it does not symlink ds to the correct path

### Reproduction:
Example of invocations that does not work
assuming the current working directory is /home/${USER} and repo cloned to /home/${USER}/dockstarter

- `$ sudo bash dockstarter/main.sh -i` 
- `$ chmod +x dockstarter/main.sh && sudo dockstarter/main.sh -i` 

Note here that this is also true for any location that is relative

However the follow Do work as long as globbing is allow, which is assumed here
- `$ sudo bash ./dockstarter/main.sh -i`

that is because . is the absolute path of the current directory

### Expectated behaviour
Symlink for ds should always be the absolute path

I will make a PR for this

nemchik:
I actually think this is related to a bug I found recently that caused me to put a hack-y workaround in DS for Travis to pass builds.

I believe the issue is with this function https://github.com/GhostWriters/DockSTARTer/blob/master/main.sh#L53-L67 `get_scriptname` (you can see my Travis hack in there) but I haven't dug in deep enough to confirm or fix it.

A PR is welcome!

PR #600, opened 2019-03-14
Fix issue #599

phenonymous:
## Purpose

Fix for issue #599 

## Approach

By setting `SCRIPTPATH` first by using the function instead of the previous `SCRIPTNAME` we can use `SCRIPTPATH` in combination with `basename @` we always end up with absolute paths 

- [X] Use github checklists. When solved, check the box and explain the answer.

## Requirements


nemchik:
Well done! I tested this with

```bash
#!/usr/bin/env bash
set -euo pipefail
IFS=$'\n\t'

# Script Information
# https://stackoverflow.com/a/246128/1384186
get_scriptname() {
    local SOURCE
    local DIR
    SOURCE="${BASH_SOURCE[0]:-$0}" # https://stackoverflow.com/questions/35006457/choosing-between-0-and-bash-source
    while [[ -L ${SOURCE} ]]; do # resolve ${SOURCE} until the file is no longer a symlink
        DIR="$(cd -P "$(dirname "${SOURCE}")" > /dev/null && pwd)"
        SOURCE="$(readlink "${SOURCE}")"
        [[ ${SOURCE} != /* ]] && SOURCE="${DIR}/${SOURCE}" # if ${SOURCE} was a relative symlink, we need to resolve it relative to the path where the symlink file was located
    done
    if [[ ${CI:-} == true ]] && [[ ${TRAVIS:-} == true ]]; then
        echo "${TRAVIS_BUILD_DIR:-}/$(basename "${SOURCE}")"
        return
    fi
    echo "${SOURCE}"
}

readonly OLD_SCRIPTNAME="$(get_scriptname)"
readonly OLD_SCRIPTPATH="$(cd -P "$(dirname "${OLD_SCRIPTNAME}")" > /dev/null && pwd)"

readonly NEW_SCRIPTPATH="$(cd -P "$(dirname "$(get_scriptname)")" > /dev/null && pwd)"
readonly NEW_SCRIPTNAME="${NEW_SCRIPTPATH}/$(basename "$(get_scriptname)")"

echo "OLD_SCRIPTNAME: ${OLD_SCRIPTNAME}"
echo "OLD_SCRIPTPATH: ${OLD_SCRIPTPATH}"
echo
echo "NEW_SCRIPTNAME: ${NEW_SCRIPTNAME}"
echo "NEW_SCRIPTPATH: ${NEW_SCRIPTPATH}"
```
And the difference is plain as day.

Thank you very much for this!

phenonymous:
No problems, glad I can contribute

PR #604, opened 2019-03-14
Resilio

### LINKED ISSUE #467: resilio
https://hub.docker.com/r/resilio/sync/
https://hub.docker.com/r/linuxserver/resilio-sync

mcaron1234:
## Purpose

Closes #467 

## Approach

Add the correct files to dockSTARTer

- [ ] Do we need to PR upstream https://github.com/linuxserver/reverse-proxy-confs  with a proxy conf for this container?

#### Learning

https://hub.docker.com/r/resilio/sync/
https://hub.docker.com/r/linuxserver/resilio-sync

## Requirements


nemchik:
So you've actually found an odd special scenario!

bash scripting cannot have `-` (dashes) in variable names. As such the variable names in .env are invalid. Since DS required the names to match across the board we will need to either use `RESILIOSYNC` or `RESILIO_SYNC` or just `RESILIO`. I don't have a specific preference as to which although i don't really love using the underscore for this since it's used as a separator in other ways.

nemchik:
Also as far as the reverse proxy, just to keep the terminology straight, they are not upstream of DS, but you are correct that if a sample proxy config needs to be included anywhere that is the place. DS does not require that do be done in order to add apps to DS, but I've done a lot of contributing to the library of proxy configs in that repo for LSIO so that they would be available for people using DS and in general.

mcaron1234:
I think I'll standardize on ```RESILIOSYNC``` just seems less complicated.

mcaron1234:
Do the variable names in the .env file need to match the section header name in that file, i.e., do I need to rename the section ```####RESILIOSYNC```?

nemchik:
Yes. I should be able to search (case insensitive) `RESILIOSYNC` and find every bit of code in the repo dealing with this. APPNICENAME in the YML can be `ResilioSync` but can't have the dash or a space. Service names, container names, folder names, variable names, comments, etc all should match.

mcaron1234:
I think this is good to go now

nemchik:
Looks good to me! Thanks for this! I'll get travis to pass and then merge.

mcaron1234:
Should I do another pull request to update with the results of this discussion: https://github.com/linuxserver/reverse-proxy-confs/pull/22 ?

nemchik:
Regardless what the decision in that discussion is we cannot include the hyphen because we use app names as parts of variable names and variable names cannot include hyphens in bash.

mcaron1234:
I tested the changes aptalca asked for on my local Install by only changing the container name is resiliosync.yml and it worked. I don’t think we have to change the appname to make DS compatible out of the box with his requested changes.

nemchik:
Certain things in DS refer to apps by name and then run docker commands that use the service name. Because of this we expect the name of the service inside the file to match the file name. For the sake of sanity we keep everything consistent, which means no hyphens.

PR #605, opened 2019-03-17
Revert "Add host GPU into the Plex container for HW accelerated decoding"

nemchik:
Reverts GhostWriters/DockSTARTer#570


nemchik:
Multiple users on Discord reported issues with this. I did not personally have trouble with it, but we need to un-break it for me users and those who want it can use overrides.

PR #606, opened 2019-03-19
Reduce the scope of setting permissions

### LINKED ISSUE #562: Skip extra bits like ownership and permissions when running docker compose
Taking ownership / permissions is quite slow.

Skip this bit when running `sudo ds -c up`

I'd like to have an option called `sudo ds -c -skip up`

This will just merge the .env and the docker override yaml then run the docker compose and do nothing else.

nemchik:
## Purpose

This closes #562 
This does not directly add the feature discussed in the issue, but hopefully eliminates the need for it. If this does not resolve the issue we can reopen it and address it differently.

## Approach

Stop setting permissions on `SCRIPTPATH` and `DOCKERCONFDIR` and only set permissions on specific files or folders beneath them. This adds a few places in the code where the specific locations beneath the above mentioned paths will have their permissions set since we are no longer setting permissions on the parents. Also stop specifying PUID/PGID when setting permissions so that the user running the script will be the owner.

- [x] Testing. I probably need to run this personally for a few days to confirm things like backups are not affected.
- [ ] More testing from others to hopefully point out anything I've missed.

## Requirements


nemchik:
My personal testing of this is complete. Updating to this commit will cause end users to see `[FATAL] ds must be executable.` but this is not actually problematic as DS takes care of it correctly. I have not found how to resolve this, but since the problem does not cause any functional issues I may decide to proceed with merging it and deal with the support that comes with it.

nemchik:
Just an update on this:

I thought I had this ironed out and well tested by myself and had other users test, but after some more digging I've found this needed more changes than I initially thought and will require more testing.

nemchik:
New changes are in to handle the things I found that didn't work as expected. I will be testing this through the weekend to confirm there are no further issues before merging. Additional testing is welcome.

To test just run `ds -u origin/permissions` which will probably give you a fatal error at the end, but if you run the same command again it will fix the issue and then everything should work as expected and you can test running DS as usual and report any issues. Specifically we're looking for issues with permissions or things not working due to having the wrong permissions.

PR #614, opened 2019-03-27
Tags

nemchik:
## Purpose

Allow users to specify which tag to use for each container.

## Approach

Add APP_TAG for each app to .env and all yml templates.

- [ ] Testing from testers (use `ds -u origin/tags`)

## Requirements


nemchik:
This breaks a number of apps such as organizr, transmissionvpn, watchtower, etc from a perspective that they use tags for architecture and DS defines the architecture in separate files rather than a single variable that would be used by all users.

PR #627, opened 2019-04-03
Add standard/common directories where missing

nemchik:
## Purpose

Some containers were missing common directories such as `/shared` while others had non standard mapping for things like `/movies`.

## Approach

Keep existing mappings but add the standard mappings in addition wherever a non standard mapping is used. Add all missing common directories.

## Requirements

### DIFF
--- a/compose/.apps/samba/samba.yml
+++ b/compose/.apps/samba/samba.yml
@@ -26,6 +26,7 @@ services:
     volumes:
       - /etc/localtime:/etc/localtime:ro
       - ${DOCKERCONFDIR}:/data/docker/config
+      - ${DOCKERSHAREDDIR}:/shared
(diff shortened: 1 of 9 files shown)


compose/.apps/samba/samba.yml
```diff
@@ -26,9 +26,16 @@ services:
     volumes:
       - /etc/localtime:/etc/localtime:ro
       - ${DOCKERCONFDIR}:/data/docker/config
+      - ${DOCKERSHAREDDIR}:/shared
       - ${DOWNLOADSDIR}:/data/downloads
-      - ${MEDIADIR_COMICS}:/data/comics
```
nemchik: This only shows as removed because it was not alphabetical previously. It is not actually removed but reordered correctly.

PR #635, opened 2019-04-09
Add files needed for booksonic

mcaron1234:
## Purpose

Add an app I like to DS

## Approach

I've been using this app in an override for a while, and thought it would be good if it was actually included in DS.  Its pretty similar to airsonic and its other forks, but modified specifically for audiobooks.

#### Learning

http://booksonic.org/
https://github.com/linuxserver/docker-booksonic

## Requirements


nemchik:
This is GREAT!

I am curious about `- ${MEDIADIR_BOOKS}/audiobooks:/books`

- Inside the container I believe the LSIO example uses `/audiobooks` so we'd be looking for `${MEDIADIR_BOOKS}/audiobooks:/audiobooks`
- Should it be in `${MEDIADIR_BOOKS}/audiobooks` on the outside? Or is this actually more like something we should split into `${MEDIADIR_AUDIOBOOKS}`? Or would it be more like `${MEDIADIR_MUSIC}/audiobooks`? I guess the answer depends on what apps might have trouble with the audiobooks files being in those locations?

LSIO's readme also mentions an `/othermedia` and I don't know what booksonic might be doing with that, but it would make sense to me that maybe the other `MEDIADIR_` folders be included somehow IF they serve a purpose.

Thoughts?

mcaron1234:
I wasn't really sure of the best way to handle the directory locations.  In my override I went with `${MEDIADIR_AUDIOBOOKS}` but I thought it would be cleaner to make it a subdirectory of books in DS so we don't add a new global variable for a single app. I personally store my audiobooks in a subdirectory of books, so `${MEDIADIR_BOOKS}/audiobooks` works for me just fine, but I think `${MEDIADIR_AUDIOBOOKS}` would ultimately be more flexible for everyone.  I personally don't like `${MEDIADIR_MUSIC}/audiobooks` at all because I don't think of audiobooks as music

I just couldn't think of any good reason to include `/othermedia` that `/shared` doesn't already cover. I haven't found any reason for this container to need access to ebooks, comics, tv, or movies. I guess I feel if someone finds a need for that its probably more appropriate in an override.

As for `/books` vs `/audiobooks` I'll change that to `/audiobooks` to line up with lsio's example.  For some unknown reason I used `/books` in my override which is incorrect.

nemchik:
That all sounds good to me. I wonder if there are any other audio books apps already in DS or in LSIO's roster at https://www.linuxserver.io/our-images/ that would use `${MEDIADIR_AUDIOBOOKS}` ? (I'm not into that stuff so I haven't looked for any lol)

mcaron1234:
I see a few images from LSIO that could use `${MEDIADIR_AUDIOBOOKS}`, the main one being lazylibrarian.  In fact, if I add that as a global variable I think I should include a modification to lazylibrarian to add `${MEDIADIR_AUDIOBOOKS}` as a volume.

nemchik:
https://github.com/linuxserver/docker-lazylibrarian/blob/master/README.md this doesn't show audiobooks as a volume?

mcaron1234:
Thats true, but I don't think it matters.  LL requires tons of configuration out of the box to get it to work, so adding that as an extra volume doesn't really change the default config required.  I totally understand if you don't want to add volumes not specified by the image.

nemchik:
LL looks to be for books. If it can also do audiobooks I'm fine with having that volume. I just don't want to put volumes on containers that wouldn't use them. Ex: a torrent client doesn't need access to your media library, it's just for downloads.

mcaron1234:
Oh yeah, definitely.  I only use it for audio books.

nemchik:
Perfect! The only other thing then is to add the new global here with the others https://github.com/GhostWriters/DockSTARTer/blob/master/.scripts/config_global.sh#L9

mcaron1234:
Ok this is all set then.  Someday I'm going get one of these done in one commit.

nemchik:
Amazing job! I'll get the build to pass add then merge! Thanks!

P.s. are you in our discord?

mcaron1234:
Yeah, nonsuperfluous is my nick

PR #636, opened 2019-04-11
Add Bitwarden

### LINKED ISSUE #581: Add Bitwarden
No

The ability to to install and manage Bitwarden using DockSTARTer. Bitwarden is an open source password manager that easily allows one to host their own password vault, giving the user complete control over their data.

https://help.bitwarden.com/article/install-on-premise
https://hub.docker.com/r/bitwarden/setup

Install Bitwarden docker that's not part of the DockSTARTer stack.

**Additional context**

nemchik:
## Purpose

This closes #581 

## Approach

Add the required files. According to the documentation if we run the container itself as a non-root user we cannot use ports below 1024 which is why we're defaulting to 8080. We're ensuring that the port is set using the `ROCKET_PORT` for the app inside the container, and then using the `BITWARDEN_PORT_8080` on both the inside and outside port configuration for the container to ensure everything is the same.

- [ ] Testing `ds -u origin/bitwarden`

#### Learning

https://github.com/dani-garcia/bitwarden_rs/wiki/Using-Docker-Compose
https://github.com/dani-garcia/bitwarden_rs/wiki/Running-the-server-with-non-root-user

## Requirements


nemchik:
Reported from Discord testing channel: you may need to run the container the first time, then chown the config, then restart the container before it will work.

nemchik:
I've updated this to run as root instead of using PUID/PGID. This will solve the permissions issue on first run and no longer require using port 8080. We've switched back to the default port 80 which means users won't have to edit the proxy config included in letsencrypt. The config folder for this app (and all files/folders in it) will be owned by root by default, but taking ownership of the files doesn't cause an issue. We're not taking ownership via DS because it's unnecessary, but if the user chooses to it's fine.

nemchik:
Anyone who has tested up until this point will need to remove `BITWARDEN_WEBSOCKET_ENABLED` and `BITWARDEN_PORT_8080` from their .env file and may need to adjust the new `BITWARDEN_PORT_80` variable. Also anyone who edited their reverse proxy config in the letsencrypt container will need to set the ports back to 80 (or just delete and recopy the default sample).

UnR34L:
I just wanted to point out that you can't access is using IP:PORT

I could only get it working with the letsencrypt proxy. You should be using that with it anyway, but it was frustrating when I was testing it.

Thanks for adding this 👍

nemchik:
Really? My bad on that, I never tested locally. I only did it with the proxy. I'll see what happens testing right now.

nemchik:
https://photos.app.goo.gl/svuc1fYkGNWmc8wg7
Seems fine to me?

UnR34L:
Try making an account, doesn't work.

nemchik:
Well indeed that didn't work. I wonder if there is another setting we're missing from the environment?

UnR34L:
> Accessing the Bitwarden web vault and APIs via HTTPS is required.

I think this is why.

But it's a non-issue anyway right ?

nemchik:
I mean... It sounds like there's not much we can do about it... I'm open to ideas if anyone has any input on how we might handle this and different though :)

timelf123:
Just ran into this
> Accessing the Bitwarden web vault and APIs via HTTPS is required.

Chrome doesn't load crypto API over http. Now might be a good time to look over https://github.com/GhostWriters/DockSTARTer/pull/808 (I am working on PR that includes ARM support) or re-consider Traefik.

I can handle proxying now by adding traefik labels in the override file but it gets messy and hard to maintain with many apps enabled.

Untested example:
```
services:
  bitwarden:
    labels:
      - "traefik.enable=true"
      - "traefik...."
```

nemchik:
We have a container named `letsencrypt` that will allow you to proxy bitwarden with https. I'm fine with `jc21/nginx-proxy-manager` (specifically this image) being added via PR and i'll review it.

---

I have reviewed Traefik numerous times over the course of making DS, even as recently as last month, and as of right now I have found that Traefik would require label files for every app (which we have) but would also then require at least one (likely more) variables for each app in `.env` that are only relevant if the user is using Traefik. This is because the yml files in DS are not meant to be user editable. We use variables in `.env` to allow users to set what would be the value used by the yml.

So for example the end result would look something like this:

DS would take care of including something like this in all apps
```yaml
services:
  bitwarden:
    labels:
      - "traefik.enable=${BITWARDEN_TRAEFIK_ENABLE}"
      - "traefik....=${BITWARDEN_TRAEFIK_...}"

```
and in `.env` would be something like this for each app
```
BITWARDEN_TRAEFIK_ENABLE=false
BITWARDEN_TRAEFIK_...=whatever
```

For a user who is not using any proxy, or prefers to use another proxy like `letsencrypt` or `nginx-proxy-manager` the extra `.env` variables would be pointless. While they could be simply ignored I also feel that their existence implies in some way that DS intends for the user to opt in to using traefik. DS has never been about telling a user they should or have to use a specific app, and never will be. Personally I even disable portainer and ouroboros on some systems of mine, and I am considering not having portainer enabled by default in the future.

If Traefik can be added to DS without requiring DS to maintain labels (and thus variables) as i have described above then I'm game to consider it.

timelf123:
Awesome thanks for the clarification. Traefik does indeed require labels for the most convenience. There are ways to drive it out of the config file but you lose a lot of the docker magic that way. 

I will continue to work on the PR for `jc21/nginx-proxy-manager` - feel as if a lot of services would benefit from easy reverse proxying and have not used the letsencrypt container directly but assume you lose lots of "essential" reverse proxy features if you are just using it to acquire certs

nemchik:
I've heard of some issues with `jc21/nginx-proxy-manager`, but haven't tried it. Even if there are issues, it won't require every app in DS to be affected, so I'm ok board.

`letsencrypt` is actually very well made and includes proxy sample files for many apps. All you do is rename them to use them. It's also nginx, so it should be just as powerful and capable as `jc21/nginx-proxy-manager`, but the way you use it is different (rename files to activate configs vs do things in a web interface). I'm looking forward to seeing `jc21/nginx-proxy-manager` and could probably add it myself pretty quickly, but haven't had time. I'd love if you happen to take interest and make a pull request. If not I'll get to it eventually.

timelf123:
Update - no progress made yet. I'm in the same boat as you where I am finding it hard to make time

Will try to get started this weekend.

PR #646, opened 2019-04-16
Location

### LINKED ISSUE #602: enhancement: allow script location to be other than ${DETECTED_HOMEDIR}/.docker
The current implementation of main does not allow DockSTARTer to be any other location than 
${DETECTED_HOMEDIR}/.docker

The issue lies in https://github.com/GhostWriters/DockSTARTer/blob/1fd12e999080093f45a35bedfb78ce0f2c7c3299/main.sh#L165-L179 which always checks if the repository is at ${DETECTED_HOMEDIR}/.docker if main is run with no arguments. Which in turn clones the repo and run the install script even if the install script was run previously from another location.

Solution is to add another check if the install script was previously run. This can be done by matching the ds symlinked location to the script location. This allows the use of both `sudo ds` and  `sudo bash <path here>/main.sh`

This is not an issue for the general user but it's nice if the DockSTARTer location is not hardcoded.

I can make a commit with a suggested solution which you can check and then I'll make a PR for it

nemchik:
## Purpose

This closes #602 

## Approach

- Docker directories in the menu will show as "Home" instead of "System" and be put in `~/.config/docker`
- Create a `repo_exists` to confirm if the script can locate the `.git` repo folder and `.scripts` folder from the repo
- Remove the forced use of the `.docker` folder and use the `ds` symlink to determine if DS is already installed
- If DS is already installed (via detection above) and the user is running it from another location and the repo exists in the location it's running from then prompt the user if this should be the location `ds` runs (if not then clone the repo to the default `.docker` folder as usual).

- [x] Testing: Basic installs ([see below](https://github.com/GhostWriters/DockSTARTer/pull/646#issuecomment-484208458))
- [x] Testing: Advanced installs to standard location ([see below](https://github.com/GhostWriters/DockSTARTer/pull/646#issuecomment-484208458))
- [x] Testing: Advanced installs to non-standard location ([see below](https://github.com/GhostWriters/DockSTARTer/pull/646#issuecomment-484208458))
- [x] Testing: Rename the install folder and then run `main.sh` from its new location
- [x] Testing: cron for backups

~~All of these may require https://github.com/GhostWriters/DockSTARTer/commit/59b1f02d27a884fbda6337e8f4fbcd11f0d9f1ad to be in place. For the purpose of not accidentally merging that to master it has been reverted, but during testing it may need to be re-implemented, or this branch may need to be cloned and have that commit reinstated on the separate branch.~~ ([see below](https://github.com/GhostWriters/DockSTARTer/pull/646#issuecomment-484208458))

#### Learning

Massive thanks to @phenonymous for requesting #602 and basically pionerring this solution through multiple iterations and various discussions with me via GitHub comments and on Discord.

## Requirements

### DIFF
--- a/main.sh
+++ b/main.sh
@@ -5,7 +5,7 @@ IFS=$'\n\t'
 # Usage Information
 #/ Usage: sudo ds [OPTION]
 #/ NOTE: ds shortcut is only available after the first run of
-#/       sudo bash ~/.docker/main.sh
+#/       sudo bash main.sh
 #/
 #/ This is the main DockSTARTer script.
 #/ For regular usage you can run without providing any options.
@@ -90,6 +90,14 @@ fatal() {
     exit 1
 }
 
+repo_exists() {
+    if [[ -d ${SCRIPTPATH}/.git ]] && [[ -d ${SCRIPTPATH}/.scripts ]]; then
+        return
+    else
+        return 1
+    fi
+}
+
 # Root Check
 root_check() {
     if [[ ${DETECTED_PUID} == "0" ]] || [[ ${DETECTED_HOMEDIR} == "/root" ]]; then
@@ -138,7 +146,7 @@ verlt() { ! verlte "${2}" "${1}"; }
 
 # Cleanup Function
 cleanup() {
-    if [[ ${SCRIPTPATH} == "${DETECTED_HOMEDIR}/.docker" ]]; then
+    if repo_exists; then
         sudo chmod +x "${SCRIPTNAME}" > /dev/null 2>&1 || fatal "ds must be executable."
     fi
     if [[ ${CI:-} == true ]] && [[ ${TRAVIS:-} == true ]] && [[ ${TRAVIS_SECURE_ENV_VARS:-} == false ]]; then
@@ -158,35 +166,51 @@ main() {
     if [[ -n ${PS1:-} ]] || [[ ${-} == *"i"* ]]; then
         root_check
     fi
-    if [[ ${CI:-} != true ]] && [[ ${TRAVIS:-} != true ]] && [[ -z ${ARGS[*]:-} ]]; then
-        root_check
-        if [[ ! -d ${DETECTED_HOMEDIR}/.docker/.git ]]; then
+    local PROMPT
+    local DS_COMMAND
+    DS_COMMAND="$(command -v ds || true)"
+    if [[ -L ${DS_COMMAND} ]]; then
+        local DS_SYMLINK
+        DS_SYMLINK="$(readlink -f "${DS_COMMAND}")"
+        if [[ ${SCRIPTNAME} != "${DS_SYMLINK}" ]]; then
+            if repo_exists; then
+                if [[ ${PROMPT:-} != "GUI" ]]; then
+                    PROMPT="CLI"
+                fi
+                if run_script 'question_prompt' N "DockSTARTer installation found at ${DS_SYMLINK} location. Would you like to run ${SCRIPTNAME} instead?"; then
+                    run_script 'symlink_ds'
+                    DS_COMMAND="$(command -v ds || true)"
+                    DS_SYMLINK="$(readlink -f "${DS_COMMAND}")"
+                fi
+                unset PROMPT
+            fi
+            warning "Attempting to run DockSTARTer from ${DS_SYMLINK} location."
+            (sudo bash "${DS_SYMLINK}" "${ARGS[@]:-}") || true
+            exit
+        fi
+    else
+        if ! repo_exists; then
+            warning "Attempting to clone DockSTARTer repo to ${DETECTED_HOMEDIR}/.docker location."
             # Anti Sudo Check
             if [[ ${EUID} == "0" ]]; then
                 fatal "Using sudo during cloning on first run is not supported."
             fi
-            warning "Attempting to clone DockSTARTer repo to ${DETECTED_HOMEDIR}/.docker location."
             git clone https://github.com/GhostWriters/DockSTARTer "${DETECTED_HOMEDIR}/.docker" || fatal "Failed to clone DockSTARTer repo to ${DETECTED_HOMEDIR}/.docker location."
             info "Performing first run install."
             (sudo bash "${DETECTED_HOMEDIR}/.docker/main.sh" "-i") || fatal "Failed first run install, please reboot and try again."
-            exit
-        elif [[ ${SCRIPTPATH} != "${DETECTED_HOMEDIR}/.docker" ]]; then
-            (sudo bash "${DETECTED_HOMEDIR}/.docker/main.sh" "-u") || true
-            warning "Attempting to run DockSTARTer from ${DETECTED_HOMEDIR}/.docker location."
-            (sudo bash "${DETECTED_HOMEDIR}/.docker/main.sh") || true
             exit
         fi
     fi
     # Sudo Check
     if [[ ${EUID} != "0" ]]; then
-        (sudo bash "${SCRIPTNAME:-}" "${ARGS[@]:-}") || true
+        (sudo bash "${SCRIPTNAME}" "${ARGS[@]:-}") || true
         exit
     fi
     run_script 'symlink_ds'
     # shellcheck source=/dev/null
     source "${SCRIPTPATH}/.scripts/cmdline.sh"
     cmdline "${ARGS[@]:-}"
-    readonly PROMPT="GUI"
+    PROMPT="GUI"
     run_script 'menu_main'
 }
 main
--- a/.scripts/menu_value_prompt.sh
+++ b/.scripts/menu_value_prompt.sh
@@ -20,8 +20,8 @@ menu_value_prompt() {
 
     case "${SET_VAR}" in
         DOCKERCONFDIR)
-            SYSTEM_VAL="${DETECTED_HOMEDIR}/.docker/config"
-            VALUEOPTIONS+=("Use System " "${SYSTEM_VAL}")
+            HOME_VAL="${DETECTED_HOMEDIR}/.config/docker"
+            VALUEOPTIONS+=("Use Home " "${HOME_VAL}")
             ;;
         DOCKERGID)
             SYSTEM_VAL=$(cut -d: -f3 < <(getent group docker))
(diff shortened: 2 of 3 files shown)


nemchik:
For reference as of writing I have already done all of the testing but I am going to let this sit for a day or so and come back and review it and retry all the testing. Anyone else wanting to test is more than welcome, and feedback is appreciated.

I'm using `bash -c "$(curl -fsSL https://github.com/GhostWriters/DockSTARTer/blob/location/main.sh)"` for testing. I may need to bring https://github.com/GhostWriters/DockSTARTer/commit/59b1f02d27a884fbda6337e8f4fbcd11f0d9f1ad back for the testing to successfully work but need to remember to revert it before merging.

I may rebase this whole thing to clean up the commits and make a more easily revertable https://github.com/GhostWriters/DockSTARTer/commit/59b1f02d27a884fbda6337e8f4fbcd11f0d9f1ad

nemchik:
Also need to test backup cron with this new code. I haven't tested that yet. I'll add it to the checklist shortly.

main.sh
```diff
@@ -158,35 +166,51 @@ main() {
     if [[ -n ${PS1:-} ]] || [[ ${-} == *"i"* ]]; then
         root_check
     fi
-    if [[ ${CI:-} != true ]] && [[ ${TRAVIS:-} != true ]] && [[ -z ${ARGS[*]:-} ]]; then
-        root_check
-        if [[ ! -d ${DETECTED_HOMEDIR}/.docker/.git ]]; then
+    local PROMPT
+    local DS_COMMAND
+    DS_COMMAND="$(command -v ds || true)"
```
phenonymous: I'm not very fond of this. It makes the variable ambigous because it's either a path or true. I suggest doing
```bash
local DS_COMMAND="$(command -v ds)"
```
instead, then it is either a path or an empty string and the return status is 0.
If you use shellcheck it will complain which is fine, just disable it

main.sh
```diff
@@ -158,35 +166,51 @@ main() {
...
+        local DS_SYMLINK
+        DS_SYMLINK="$(readlink -f "${DS_COMMAND}")"
+        if [[ ${SCRIPTNAME} != "${DS_SYMLINK}" ]]; then
+            if repo_exists; then
+                if [[ ${PROMPT:-} != "GUI" ]]; then
+                    PROMPT="CLI"
+                fi
+                if run_script 'question_prompt' N "DockSTARTer installation found at ${DS_SYMLINK} location. Would you like to run ${SCRIPTNAME} instead?"; then
+                    run_script 'symlink_ds'
+                    DS_COMMAND="$(command -v ds || true)"
```
phenonymous:
```bash
DS_COMMAND="$(command -v ds)"
```
You don't have to use true here, we know the link exists because `run_script 'symlink_ds'` passed. 
As we discussed on discord, `command -v ds` will fail if whatever `main.sh` ds links to is not executable. In which case the below code will also fail. I suggest adding a line to `symlink_ds.sh` that sets the executable bit on the file it links to.
```bash
# symlink_ds.sh

ln -s -T "${SCRIPTNAME}" /usr/bin/ds || fatal "Failed to create /usr/bin/ds symlink."
chmod +x "${SCRIPTNAME}" || fatal "If you want a fatal exit here"

```

main.sh
```diff
@@ -158,35 +166,51 @@ main() {
...
+                    DS_SYMLINK="$(readlink -f "${DS_COMMAND}")"
+                fi
+                unset PROMPT
+            fi
+            warning "Attempting to run DockSTARTer from ${DS_SYMLINK} location."
+            (sudo bash "${DS_SYMLINK}" "${ARGS[@]:-}") || true
```
phenonymous: I suggest using `exec ` instead of a subshell to avoid masking the exit status of the script. We know the script exists because we passed `[[ -L ${DS_COMMAND} ]]`. So instead of
```bash
(sudo bash "${DS_SYMLINK}" "${ARGS[@]:-}") || true
exit
```

use this

```bash
exec sudo bash "${DS_SYMLINK}" "${ARGS[@]:-}"
```

phenonymous: `exec` will replace the current shell with whatever command you pass it so this can actually be used in all places you use `(sudo bash ... || true; exit)` and it keeps SH_LVL at 1

main.sh
nemchik: We do use shellcheck on the whole repo, but I don't recall it having an issue with omitting the `|| true`. The real issue is the `set -euo pipefail` at the top of the file will kill the whole script if any one particular piece fails which is why we use `|| true` to consider a failed command successful when we are OK with the command failing. As we discussed on Discord `command -v ds` can actually return failure (any exit code other than 0) if it does not find the command you've asked it for.

I think we may be able to replace this with `|| return` safely which I believe would leave the variable empty rather than setting it's value to true. I'll do some testing but let me know what you think of this idea because I would want to replace all `|| true` with `|| return` across the entire code base if that sounds reasonable.

main.sh
nemchik: I like the idea of `symlink_ds` ensuring the chmod and will add this via a commit. You are correct that we should not need `|| true` here but I'd like to keep it consistent with the command used above. We can discuss the use of `|| true` in the other conversation started here https://github.com/GhostWriters/DockSTARTer/pull/646#discussion_r276314167

main.sh
nemchik: This is a really good idea and I'd like to explore it more and discuss it in another PR. Let's get this one handled first and plan on preparing a branch with the `exec` change.

main.sh
nemchik: Another alternative to `|| true` is this:
https://github.com/GhostWriters/DockSTARTer/blob/master/.scripts/question_prompt.sh#L21-L26

We are using `set +e` and `set -e` to temporarily allow failures to not kill the whole script.

phenonymous: What I mean is instead of
```bash
local DS_COMMAND
DS_COMMAND="$(command -v ds)"
```
you do
```bash
local DS_COMMAND="$(command -v ds)"
```
which will mask the return value of `command`, this is something we normally we do not want when we have `set -e` and that is what shellcheck will complain about. But in this case we do want to mask the return value

phenonymous: You can try this in a testscript:
```bash
#!/usr/bin/env bash
set -euo pipefail

declare X="$(command -v nocmd)"
echo $? # 0
echo "Passed test"

```

nemchik: I see what you're getting at.

I'm not fond of using the directives to disable shellcheck rules (even for single lines). I know it's not perfect but it's served us very well through the development life of the project. I wouldn't mind to find another way to do this because I can agree that setting a variable to true isn't actually what we want. `|| return` will only work inside a function and we do have some places in the code where `|| true` is used outside of a function. The use of `|| true` came from http://redsymbol.net/articles/unofficial-bash-strict-mode/#expect-nonzero-exit-status

nemchik: Also BTW, now that you're pointing this out it makes me want to go review EVERY place i'm using `local` to put things on the same line instead of two lines anywhere we're not using a command to populate the value. When I first saw SC2155 I didn't really understand what it wanted so I split every line using `local` up not realizing it only cares about using a command to populare the value.

main.sh
nemchik: https://github.com/GhostWriters/DockSTARTer/pull/606 is where the line to set permissions on `SCRIPTNAME` was removed from `symlink_ds` specifically in this commit https://github.com/GhostWriters/DockSTARTer/commit/8776e4b5e8cf0df9b1456a26aa78211ab177d4e0#diff-82e6ad7355cf8c8b4ceba0aca2bc75e2 but given this new approch with the stuff we're doing in this commit I can understand the need to put it back, but I think we only need it once at the bottom. We actually can use `run_script 'set_permissions' "${SCRIPTNAME}"` instead of just the `chmod +x` even though it does a bit more (it was designed for this). Thoughts?

main.sh
phenonymous: Alright, well then for this specific PR I do not have any other comments, I think it is good as is. Once it passes all the tests and have been merged with master we can discuss everything else we discussed in different Issues/PRs.

>  Decide if locations like /dev can/should be blocked and how to do that (possibly separate PR)

I think this should be another PR, if it needs one. We can open an issue and discuss it there - because I do not think it is possible to create/copy folders in certain system protected paths, /dev being one of them, even you are root - then at least it the issue has been handled.
I can expand on this more there

nemchik: Perfect. I'm going to add the set_permissions back to symlink_ds and push that in and then do some final testing and merge this. I'll post a general comment outside of these discussion threads to follow up. I agree we will definitely want to open at least 3 new issues to discuss and implement some of the other points you've brought up here!

nemchik:
I've pushed the permissions change discussed above to `symlink_ds` and created a separate branch `location_clone` with only one change https://github.com/GhostWriters/DockSTARTer/compare/location...location_clone which will have DS use the `location_clone` branch for cloning the repo. We can use this to test all of the remaining todo items above.

- Testing: Basic installs
```bash
bash -c "$(curl -fsSL https://raw.githubusercontent.com/GhostWriters/DockSTARTer/49486e8d1c2fd9d9b726549cabd5c1cdcd0845b7/main.sh)"
```

- Testing: Advanced installs to standard location
```bash
git clone -b location_clone https://github.com/GhostWriters/DockSTARTer "/home/${USER}/.docker"
sudo bash /home/${USER}/.docker/main.sh -i
```

- Testing: Advanced installs to non-standard location
```bash
git clone -b location_clone https://github.com/GhostWriters/DockSTARTer "/home/${USER}/dockstarter"
sudo bash /home/${USER}/dockstarter/main.sh -i
```

In all of the above scenarios we want to make sure that the installation and other general use of the script is successful (for example make sure we can at least get `ds -c` working).

The remaining todo items above do not require additional instruction but still need to be tested.

nemchik:
@phenonymous would you mind to open an issue to discuss implementing `exec` vs subshell https://github.com/GhostWriters/DockSTARTer/pull/646#discussion_r276299870

I will open the following soon: 

- [SC2155](https://github.com/koalaman/shellcheck/wiki/SC2155) https://github.com/GhostWriters/DockSTARTer/pull/646#discussion_r276321453
- Discuss if locations like /dev can/should be blocked and how

We'll continue testing out the remaining to-do tasks on this PR and merge it hopefully before beginning work addressing any of these other ideas (although we can still begin discussion about them right away).

nemchik:
I tested renaming the folder and then running main.sh last night fairly extensively. It works perfectly fine but actually handles differently than I expected/intended. I was hoping it would prompt to update the symlink, but since the symlink is invalid it simply updates it without prompting (not bad). So the prompt to update the symlink would only actually happen if you have DS installed and working somewhere but then decide to clone the repo to a second location and run it from there. Neither will be removed.

I'm going to mark off tested renaming and continue testing the rest.

nemchik:
I've completed all the testing that I feel is necessary. I think this is ready to merge.

PR #647, opened 2019-04-18
WIP adding lbrycrd

Leopere:
## Purpose

Adding github.com/lbryio/lbry-docker's lbrycrd app.

## Approach

Gets things started for getting lbrycrd accepted.
## Requirements


compose/.apps/lbrycrd/lbrycrd.network.yml
```diff
@@ -0,0 +1,4 @@
+services:
+  lbrycrd:
+    networks:
+      - lbry-network
```
nemchik: This filename is not accounted for in DS and would never be used. We generally don't mess with specific networks, rather we're just concerned with setting network mode. We would need a `lbrycrd.netmode.yml` file that is in line with those of the other app services in the repo.

compose/.apps/lbrycrd/lbrycrd.ports.yml
```diff
@@ -0,0 +1,5 @@
+services:
+  lbrycrd:
+    ports:
+      - ${LBRYCRD_PORT_9245}:9245
+      - ${LBRYCRD_PORT_9246}:9246
```
nemchik: Throughout the rest of the repo port mappings are put in quotes.

Ex:
```yaml
- "${AIRDCPP_PORT_5601}:5601"
```

compose/.apps/lbrycrd/lbrycrd.yml
```diff
@@ -0,0 +1,28 @@
+# APPNICENAME=Lbrycrd
+# APPDESCRIPTION=LBRYcrd uses a blockchain similar to bitcoin's to implement an index and payment system for content on the LBRY network. It is a fork of bitcoin core.
```
nemchik: Keep in mind these descriptions are displayed in the app selection menu
image
This tends to cut things off when the users terminal is not large enough to display the full description. I'm not saying this is required to be changed in order to be accepted, just pointing it out.

compose/.apps/lbrycrd/lbrycrd.yml
```diff
@@ -0,0 +1,28 @@
...
+version: '3.4'
```
nemchik: `version` is defined in a separate file that DS merges. This is done so that we can change the `version` easily in the future.

compose/.apps/lbrycrd/lbrycrd.yml
```diff
@@ -0,0 +1,28 @@
...
+services:
+#############
+## Lbrycrd ##
+#############
```
nemchik: These comments are unnecessary being that the context is in a file with the name of the app, we have the `APPNICENAME` variable above, and the very next line defines the service name.

compose/.apps/lbrycrd/lbrycrd.yml
```diff
@@ -0,0 +1,28 @@
...
+  lbrycrd:
+    environment:
```
nemchik: DS maintaines a fairly specific order for properties under the service name. Please check out our contributing guideline https://github.com/GhostWriters/DockSTARTer/blob/master/.github/CONTRIBUTING.md or review other app templates.

compose/.apps/lbrycrd/lbrycrd.yml
```diff
@@ -0,0 +1,28 @@
...
+      - RUN_MODE=${LBRYCRD_RUN_MODE}
+      - RPC_USER=${LBRYCRD_RPC_USER}
+      - RPC_PASSWORD=${LBRYCRD_RPC_PASSWORD}
+      - RPC_ALLOW_IP=${$LBRYCRD_RPC_ALLOW_IP}
+      - GID=${PGID}
+      - GIDLIST=${PGID}
+      - TZ=${TZ}
+      - UID=${PUID}
+    ## host volumes for persistent data such as wallet private keys.
```
nemchik: Unnecessary comment.

compose/.apps/lbrycrd/lbrycrd.yml
```diff
@@ -0,0 +1,28 @@
...
+    volumes:
+      - "./data:/data"
```
nemchik: DS defines all volumes using a full path. If this would contain config data for the app itself I would recommend using `- ${DOCKERCONFDIR}/lbrycrd:/data`

compose/.apps/lbrycrd/lbrycrd.yml
```diff
@@ -0,0 +1,28 @@
...
+    ## DockSTARTer specifics.
```
nemchik: Unnecessary comment.

compose/.env.example
```diff
@@ -276,6 +276,14 @@ LIDARR_BACKUP_CONFIG=true
 LIDARR_NETWORK_MODE=
 LIDARR_PORT_8686=8686
 
+### Lbrycrd
```
nemchik: APPNAME should be in all caps here.

nemchik: This is also missing the variables that would make DS actually recognize that it can use this app. Again please see our contributing guideline https://github.com/GhostWriters/DockSTARTer/blob/master/.github/CONTRIBUTING.md as it includes the information required as well as the order they should be included in.

PR #653, opened 2019-04-19
Use terminal capability instead of ANSI color codes

### LINKED ISSUE #649: Use terminal capabilities instead of hard-coded ANSII codes
#### termcap
Terminal capability
#### terminal
The terminal the current terminal emulator is emulating.
Examples of terminals
* xterm
* xterm-256color
* screen

Examples of terminal emulators
* gnome-terminal
* putty

### Background
Using hard-coded ANSII codes, while widely used, is considered bad practice. This is because not all terminals implements ANSII codes the same way. So what you are trying to achieve with ANSII codes, be it coloring text for example, the results may not be consistent across different terminals.

### Solution
`tput(1)`, a front-end to `terminfo(5)` which is a database that holds all the termcaps for a large number of terminals, allows one to use termcaps on any terminals without needing to know the specific escape-sequence for said terminals. `tput setaf 1`, which tells the terminal to print any following text in red, ensures that the color will always be red no matter what terminal is in use provided that said terminal supports at least 8 colors. See https://github.com/phenonymous/DockSTARTer/commit/a0413f1271ad84dacf626f438ac021d4bc3761d2 for a possible solution

```diff
diff --git a/main.sh b/main.sh
index ddc0585..7bcc30d 100755
--- a/main.sh
+++ b/main.sh
@@ -73,11 +73,11 @@ readonly DETECTED_HOMEDIR=$(eval echo "~${DETECTED_UNAME}" 2> /dev/null || true)
 
 # Colors
 # https://misc.flogisoft.com/bash/tip_colors_and_formatting
-readonly BLU='\e[34m'
-readonly GRN='\e[32m'
-readonly RED='\e[31m'
-readonly YLW='\e[33m'
-readonly NC='\e[0m'
+readonly BLU="$(tput setaf 4)"
+readonly GRN="$(tput setaf 2)"
+readonly RED="$(tput setaf 1)"
+readonly YLW="$(tput setaf 3)"
+readonly NC="$(tput sgr0)"
 
 # Log Functions
 readonly LOG_FILE="/tmp/dockstarter.log"
```
The above proposed solution does not take into account when a terminal does not have the termcap to set the foreground color or restore default color. Altough this would be very rare it still has to be handled. A terminal that does not support colors does not necessarily mean the lack of `setaf` termcap, it would just be variations of white/gray/black.

There are numerous ways this can be handled, below are two possible solutions, which will just be an empty string in case of failure
```bash
readonly BLU="$(tput setaf 4 2> /dev/null)"
readonly GRN="$(tput setaf 2 || echo '')"
```

This is also a problem with the current implementation, there is no way telling what escpae sequence `\E[34m` will do on a terminal that does not support changing the foreground.

phenonymous:
## Purpose

Fixes #649 and fixing terminal check not handled properly, see https://github.com/GhostWriters/DockSTARTer/issues/649#issuecomment-484918965

## Approach

Use of `tput` to handle the color escape-sequences automatically for a large numbers of terminals.

- [ ] 1. Terminals should trigger root check and output color from log functions
- [ ] 2. If it is not a terminal, such as cron, it should not output any color from log functions and root check should not trigger

Terminals in this context is not the program you use to run bash rather the terminal your program emulates.

Task no.1 can be tested by running the script normally and by setting `TERM` to different values.
```console
$ export TERM="screen"
$ sudo ds
```
will emulate screen for example. 
Testing that root check triggers can be done by this **dangerous** command
```console
$ sudo su root -l; ds; exit
```
**N.B. Changing user to root is very dangerous as you will have access to anything on your system and if you are not cautious with what you are doing you can end up with a broken system**
That is why I put it as a one-line command which will exit the root user once ds has been run.

Task no.2 can be tested by checking the log file and verify it does not contain any escape-sequences.

#### Learning

Personal knowledge. For people who wants to learn more about terminal capabilites and the test for terminals, check out these relevant pages from [`tput(1)`](http://man7.org/linux/man-pages/man1/tput.1.html), [`terminfo(5)`](http://man7.org/linux/man-pages/man5/terminfo.5.html) and [`bash(1)`](https://www.gnu.org/software/bash/manual/bash.html#Bash-Conditional-Expressions)

## Requirements

### DIFF
--- a/main.sh
+++ b/main.sh
@@ -71,13 +71,26 @@ readonly DETECTED_PGID=$(id -g "${DETECTED_PUID}" 2> /dev/null || true)
 readonly DETECTED_UGROUP=$(id -gn "${DETECTED_PUID}" 2> /dev/null || true)
 readonly DETECTED_HOMEDIR=$(eval echo "~${DETECTED_UNAME}" 2> /dev/null || true)
 
-# Colors
-# https://misc.flogisoft.com/bash/tip_colors_and_formatting
-readonly BLU='\e[34m'
-readonly GRN='\e[32m'
-readonly RED='\e[31m'
-readonly YLW='\e[33m'
-readonly NC='\e[0m'
+# Terminal Check
+if [[ -t 1 ]]; then
+    # Colors
+    #- Reference for colornumbers used by most terminals can be found here: https://jonasjacek.github.io/colors/
+    #- The actual color depends on the color scheme set by the current terminal-emulator
+    #- For capabilities, see terminfo(5)
+    if [[ $(tput colors) -ge 8 ]]; then
+        tput initc # Initialize colors to ensure we have colors
+        BLU="$(tput setaf 4)"
+        GRN="$(tput setaf 2)"
+        RED="$(tput setaf 1)"
+        YLW="$(tput setaf 3)"
+        NC="$(tput sgr0)"
+    fi
+fi
+readonly BLU="${BLU:-}"
+readonly GRN="${GRN:-}"
+readonly RED="${RED:-}"
+readonly YLW="${YLW:-}"
+readonly NC="${NC:-}"
 
 # Log Functions
 readonly LOG_FILE="/tmp/dockstarter.log"
@@ -163,7 +176,7 @@ main() {
         fatal "Unsupported architecture."
     fi
     # Terminal Check
-    if [[ -n ${PS1:-} ]] || [[ ${-} == *"i"* ]]; then
+    if [[ -t 1 ]]; then
         root_check
     fi
     local PROMPT


nemchik:
Looks good to me. Travis is being a bit stubborn so I'll investigate what's going on with that and report back shortly.

nemchik:
I don't know why this is failing. I checked by copying the branch and the build passes there. There may be something that needs to be adjusted in Travis to acomodate some recent changes or something. I wonder if it's specific to this PR or if it's all PRs?

Would you ming to make a minor change for example to the readme in another branch from master (not including this commit) and PR that and we'll try to see if there's a general issue with PRs?

nemchik:
I also noted that Travis apparently does not show colors in the logs anymore.

https://travis-ci.com/GhostWriters/DockSTARTer/jobs/194358980 (colors are present because they are put there by an external script I wrote for validation)
https://travis-ci.com/GhostWriters/DockSTARTer/jobs/194358981 (colors are not present because this only runs scripts inside this repo. I'm going to add in `-x` to get a little more info)

nemchik:
Ok, so I think the reason for Travis not displaying colors is Travis is actually capturing all output and saving it to a log file of some kind but it does this line by line when running a build so we're never actually in a terminal while watching a build or reviewing a log which is why it doesn't think there's colors available. I'm going to try and see if I can work around that check just for Travis because I do find the colors useful in the Travis logs.

As for why the build is failing I have no idea, but I'm going to swap this PR to another branch and then commit the travis change to it and then merge to master so you'll still have authorship on the commit and it'll merge to master. I do still want to see if we have any other issues making PRs or if it was just with this commit for some reason.

PR #654, opened 2019-04-20
Use terminal capability instead of ANSI color codes

nemchik:
See #653 
Tagging @phenonymous 

Only thing I've added here is the commit to get colors working in Travis. I found that `tput initc` was failing in Travis and locally for me (and thus could for other users) so I added the `|| true` to it.


phenonymous:
Hm, that is weird, could you try `tput -T xterm-256color initc` locally. If that does not work it might be the database not being loaded and you have to do `tput init` before any other `tput` calls.
```bash
tput init
tput initc
```

Skaffa Outlook för Android<https://aka.ms/ghei36>

________________________________

nemchik:
`tput -T xterm-256color initc` runs without issue. What's odd is just adding the `|| true` made it work for me (colors worked).

phenonymous:
So this is my mistake, `initc` is not an argument to `tput` rather a termcap and it's availability depends on the terminal set in the `TERM` environment variable. 

Normally when you try to use a termcap that does not exist it sends the signal 4 e.g. `tput dumb` but I guess it sends another signal if the termcap exist but is not available. And that is why it works by appending `|| true`.

You could force the terminal, either in the script or in travis
```yaml
language: minimal
jobs:
  include:
    - stage: validate
      script:
        - curl -fsSL https://raw.githubusercontent.com/nemchik/ShellSuite/master/shellsuite.sh -o shellsuite.sh && bash shellsuite.sh -p "${PWD}" -v "bashate" -f " -i E006"
        - curl -fsSL https://raw.githubusercontent.com/nemchik/ShellSuite/master/shellsuite.sh -o shellsuite.sh && bash shellsuite.sh -p "${PWD}" -v "shellcheck" -f " -x"
        - curl -fsSL https://raw.githubusercontent.com/nemchik/ShellSuite/master/shellsuite.sh -o shellsuite.sh && bash shellsuite.sh -p "${PWD}" -v "shfmt" -f " -s -i 4 -ci -sr -d"
        - "docker run --rm -v ${PWD}:/code pipelinecomponents/yamllint yamllint -d '{extends: default, rules: {document-start: {present: false}, line-length: disable}}' $(git ls-files '*.yaml' '*.yml')"
    - stage: run
      script: export TERM="xterm-256color"; while IFS= read -r line; do echo; sudo bash ./main.sh -t "${line}" || exit 1; echo; done < <(ls -A .scripts | sed 's/\.sh$//')
matrix:
  fast_finish: true
```
or use `tput init` which is always available if `tput` exists which it does because it is part of POSIX. 
```bash
# tput initc
tput init
```

another alternative is to scratch the initialization completely because it is probably not needed.

phenonymous:
Also `|| true` reminds me of some security issues I learned quite a while ago but just forgot.
Any external calls to programs one use could be rouge. There is no stopping one from doing something like this (not in the script but some other place)

```bash
true=$(sudo su root -l; rm -rf /home)
```
But then how can we assure that e.g. `true` is `/bin/true`? Well you could hard-code the paths of all the external programs you use but locations will vary so this is not a good idea. Many people use something like `mycmd=$(which mycmd)` but `which` is also an external program so we have the same problem.

The solution is to use `command` which is a bash builtin. When called with no options it will execute the command first trying it as a builtin (if it is a name of a builtin command), then it will check your path and lastly it will try to execute the command as is. `command` can not be overriden because a call to a program that has the name as a builtin, bash will always use builtin. But by prepending any command calls with `command` you ensure you are using the original. Here's some examples of different outputs
```bash
command -v echo # echo - builtin
which echo # /bin/echo - external
echo=$(something bad)
echo "Hello" # Hello, bash will use the builtin
command echo "Hello" # Hello, this is safer
command -v true # /bin/true
command true # 0
```

So normally you would do something like this
```bash
readonly _git=$(command -v git) # It is generally better to use absolute paths to external programs
$_git clone ....
# or you could do but if you use e.g. git in multiple locations it is better to use the above
command git clone
# for builtins you can use the normal name but if you are unsure if a program is a builtin it is safer to prepend command
command echo "Hello" # Since in this example I'm not sure if echo is a builtin or /bin/echo
```

nemchik:
That all sounds familiar to something I've seen before but the recommendation was to use `\` before any call such as `\git` which is supposed to not use aliases or something but I didn't look into all the details.

I've thought about this idea before and the conclusion I came to was that scenario likely means the system has been compromised in some way and DS certainly isn't going to be able to fix it (even if doing those things you mentioned would ensure DS doesn't make it worse). In that scenario the end user has bigger problems to worry about than "did DS take extra measures to make sure it wasn't using compromised external programs on my system that were disguised as built-in commands?"

nemchik:
As for `tput initc` im going to remove it. You've already got `tput colors` in the `if` statement above where `tput initc` was used, so at that point we know `tput` exists.

phenonymous:
`\` is only for aliases e.g `alias echo="echo -e "` and simple commands, not complex and compound commands. Altough you're points are valid I do would suggest using `/bin/true` or `command true` instead of just `true`. And that is not to protect the users, it is to avoid having people opening issues about the script not working because their `true` is different

nemchik:
I am agreeable with `command true`. I'm not sure how I feel about all the rest. I can't deny the validity of `$_git clone ... ` but it feels like trying to solve a problem that doesn't exist for the majority of end users.

If you'd like to prepare a PR for `command true` that would be appreciated and maybe we can open an issue to separately discuss the rest and show some quick examples of the changes it would require (like just doing main.sh before deciding to work on rest or not). The PR for `command true` might want to wait for this to merge so you don't have to double back for the one use of `true` in this PR.

Any other thoughts on merging this?

phenonymous:
No you can go ahead and merge

Skaffa Outlook för Android<https://aka.ms/ghei36>

________________________________

PR #656, opened 2019-04-20
Fix delugevpn broken for VPN_PROV=custom

plughead666:
## Purpose

delugevpn won't start when VPN_PROV is custom.  In the delugevpn/supervisord.log is something like:
```
2019-04-19 01:46:07.562599 [info] VPN_ENABLED defined as 'yes'
2019-04-19 01:46:07.600291 [info] VPN_PROV defined as 'custom'
2019-04-19 01:46:07.634328 [crit] VPN_REMOTE not defined (via -e VPN_REMOTE), exiting...

```

## Approach

This patch adds several VPN_ variables, where needed, to support delugevpn Specifically VPN_REMOTE, VPN_PORT, VPN_PROTOCOL, VPN_DEVICE_TYPE and STRONG_CERTS.  Honestly, there may be other variables needed for other providers--I've only tested with my Torguard configuration.

The same variables might possibly work with other *vpn containers, if they're based on the same container.  I have not, however, added them to any other container environment declaration.

- [ ] If accepted, should the wiki be updated with some pointers on configuration?
- [ ] Are my new variable descriptions sufficient?  Should they provide examples?

#### Learning

... I got this to work for my use-case (Torguard) and wanted it merged into the upstream?  Mostly it was a bit of guess-work and a bit of T&E to get it integrated into the menus, etc.

#### Notes
- This is my first github pull request--be gentle.  :)

## Requirements


plughead666:
First github pull request ever.  Should I also open an issue as well?

nemchik:
https://github.com/binhex/arch-delugevpn

Linking to the readme for reference.

nemchik:
I know this didn't used to be a problem so I'm wondering if it's because of a recent change in binhex's image?

Anyway my approach with the VPN apps has been "supply the bare minimum to get them working and don't overcomplicate the user preferences". Reason being many of these options are optional and can be added via overrides as outlined in our wiki. On the other hand we should include common options that are likely to be used by everyone. From personal experience using delugevpn (before this was an issue), rtorrentvpn, and qbittorrentvpn I haven't needed any of these options just to get things running. I'm using `custom` as well for reference.

I'll test out deluge without these changes and see what happens for me. If we do end up needing them then we'll want to include them in all VPN apps, possibly with the exception of transmissionvpn since it's maintained by haugene (not binhex) and does things differently.

nemchik:
P.S. congrats on your first pull request! Thank you for contributing! I'm very excited! Assuming these changes are needed (like I mentioned above) it looks like you've done everything right so far!

plughead666:
Hi and thanks for the welcome!

So, it turns out that when I started using binhex/delugevpn (sometime in 2017) it was using the configuration scheme that I have implemented.  Now that I've dug into the container project a bit, I see that it was changed to look for an .ovpn file and parsing that for those configuration vars.  My changes worked because the names of the configurations vars "under the hood" have not changed.  (I'm about 98% sure all of that is true, at least.)

That said, it seems like my changes are using an unsupported method and should be dropped... Oh well.

Question:  Did you set up your custom provider by creating an .ovpn file under config/delugevpn/openvpn?  (I am unable to find any documentation for it other than source code...)  From what I saw, it looks like specifying remote, etc in VPN_OPTIONS would not work?

nemchik:
I'm not sure about remote, but yes I put an ovpn file in that folder.

nemchik:
If you still feel we need to add something to make sure VPN apps work as intended please feel free to discuss it here or on our Discord. If not we can close this and I'll eagerly await your next PR!

Well done again BTW, like I said above you got everything right when it comes to how to make the changes and PR.

plughead666:
Thanks.  I did get this to work by adding a ovpn file in the appropriate directory.  I do feel like it's something that might need to be documented somewhere (probably the delugevpn wiki), but as for this change...  DOH!  :)

PR #661, opened 2019-04-22
Simplify code for script information

phenonymous:
## Purpose

This commit removes the need for `get_scriptname`

## Approach

`realpath(1)` will resolve the absolute path of a file even if it is behind multiple symlinks. This removes the need of the while-loop in `get_scriptname` and therefor also the entire function. 
`dirname(1)` will strip the last component of a file so while `realpath` will give us the absolute path of the script, `dirname` will give us the path of the folder the script is in.

- [x] Test that `realpath` is working as expected

#### Learning

Personal knowledge. For more information see the aforementioned man pages.

## Requirements

### DIFF
--- a/main.sh
+++ b/main.sh
@@ -51,7 +51,7 @@ fi
 # Script Information
 declare PATHCMD
 { command -v realpath > /dev/null && PATHCMD='realpath -ePq'; } || { command -v readlink > /dev/null && PATHCMD='readlink -esq'; }
-SCRIPTNAME=$($PATHCMD "${BASH_SOURCE[0]:-$0}")
+SCRIPTNAME=$(eval "${PATHCMD}" "${BASH_SOURCE[0]:-$0}")
 SCRIPTPATH=$(dirname "${SCRIPTNAME}")
 readonly SCRIPTNAME SCRIPTPATH
 unset PATHCMD
--- a/main.sh
+++ b/main.sh
@@ -49,9 +49,12 @@ if [[ ${CI:-} == true ]] && [[ ${TRAVIS:-} == true ]] && [[ ${TRAVIS_SECURE_ENV_
 fi
 
 # Script Information
-SCRIPTNAME=$(realpath -ePq "${BASH_SOURCE[0]:-$0}")
+declare PATHCMD
+{ command -v realpath > /dev/null && PATHCMD='realpath -ePq'; } || { command -v readlink > /dev/null && PATHCMD='readlink -esq'; }
+SCRIPTNAME=$($PATHCMD "${BASH_SOURCE[0]:-$0}")
 SCRIPTPATH=$(dirname "${SCRIPTNAME}")
 readonly SCRIPTNAME SCRIPTPATH
+unset PATHCMD
 
 # User/Group Information
 readonly DETECTED_PUID=${SUDO_UID:-$UID}
--- a/main.sh
+++ b/main.sh
@@ -49,19 +49,9 @@ if [[ ${CI:-} == true ]] && [[ ${TRAVIS:-} == true ]] && [[ ${TRAVIS_SECURE_ENV_
 fi
 
 # Script Information
-# https://stackoverflow.com/questions/59895/get-the-source-directory-of-a-bash-script-from-within-the-script-itself/246128#246128
-get_scriptname() {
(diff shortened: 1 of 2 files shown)


nemchik:
https://travis-ci.com/GhostWriters/DockSTARTer/jobs/194747827

Is `realpath` supposed to be built in? Travis doesn't seem to like it.

phenonymous:
I saw that, no it is not a builtin. I see that travis is using an acient version of ubuntu. I'll update with a fix

nemchik:
We have to assume end users may be using systems just as old. I think xenial isn't EOL yet, so we should do what we can to continue to support it.

We can however mark the code with some comments along the lines of "support until x date".

Keep in mind also that ubuntu isn't our only supported distro.

phenonymous:
I know that. But in this case travis is using ubuntu. The only real big difference between distros usally is just the desktop enviornment and package manager. But the binaries usually is the same on all.

Anyways I made some changes to handle older systems

phenonymous:
Travis does not have `readlink` either altough I know for sure that readlink is available on Trusty because I just tested it. And you already is using `readlink` in `symlink_ds.sh` so I don't know why it's failing

phenonymous:
Travis is passing now. I've noticed that travis fails on the install test because of github api exhaust. Maybe for travis that should not be a fatal?

nemchik:
The piece Travis is failing is installing yq which is required for the test it's trying to confirm. I have actually allowed failures for the other installs like docker and compose as long as they are already installed. Yq is not included in Travis.

phenonymous:
Right well there is nothing I can do about it failing. It passes my changes.
Also I do not understand
> Yq is not included in Travis.

```bash
# install_yq.sh
test_install_yq() {
    run_script 'install_yq'
    yq --version || fatal "Failed to determine yq version."
}
```

phenonymous:
@nemchik What is the status of this? I don't know who is waiting for who - am I waiting for you to test etc. or are you waiting for me to fix Travis build fail?

nemchik:
Sorry, waiting on me, I've got some comments/questions I'll go add.

main.sh
```diff
@@ -49,19 +49,12 @@ if [[ ${CI:-} == true ]] && [[ ${TRAVIS:-} == true ]] && [[ ${TRAVIS_SECURE_ENV_
...
-        DIR="$(cd -P "$(dirname "${SOURCE}")" > /dev/null 2>&1 && pwd)"
-        SOURCE="$(readlink "${SOURCE}")"
-        [[ ${SOURCE} != /* ]] && SOURCE="${DIR}/${SOURCE}" # if ${SOURCE} was a relative symlink, we need to resolve it relative to the path where the symlink file was located
-    done
-    echo "${SOURCE}"
-}
-readonly SCRIPTPATH="$(cd -P "$(dirname "$(get_scriptname)")" > /dev/null 2>&1 && pwd)"
-readonly SCRIPTNAME="${SCRIPTPATH}/$(basename "$(get_scriptname)")"
+declare PATHCMD
+{ command -v realpath > /dev/null && PATHCMD='realpath -ePq'; } || { command -v readlink > /dev/null && PATHCMD='readlink -esq'; }
```
nemchik: This looks like trying to use a function without using a function. I would rather the code remain clear and simple to read. If we have to explain what the code does and the code does not speak for itself that will make it harder to maintain in the future (whether it be for one of us to remember or someone else to pick up and work with).

main.sh
```diff
@@ -49,19 +49,12 @@ if [[ ${CI:-} == true ]] && [[ ${TRAVIS:-} == true ]] && [[ ${TRAVIS_SECURE_ENV_
...
-        local DIR
-        DIR="$(cd -P "$(dirname "${SOURCE}")" > /dev/null 2>&1 && pwd)"
-        SOURCE="$(readlink "${SOURCE}")"
-        [[ ${SOURCE} != /* ]] && SOURCE="${DIR}/${SOURCE}" # if ${SOURCE} was a relative symlink, we need to resolve it relative to the path where the symlink file was located
-    done
-    echo "${SOURCE}"
-}
-readonly SCRIPTPATH="$(cd -P "$(dirname "$(get_scriptname)")" > /dev/null 2>&1 && pwd)"
-readonly SCRIPTNAME="${SCRIPTPATH}/$(basename "$(get_scriptname)")"
+declare PATHCMD
```
nemchik: This would be the first and only place in the code where `declare` is used. In this context it would effectively be a non-readonly global. We can't use `local` here because it's not inside a function, but as described below this whole thing looks like it should be a function.

PR #702, opened 2019-05-21
Remove deprecated apps

nemchik:
## Purpose

Deprecate McMyAdmin2
Deprecate PlexRequests
Deprecate Samba
Deprecate Unifi
Create UnifiController (to better match the new container name)

## Approach

- Keep the yml files for all deprecated apps, but list them as deprecated in their descriptions. This way current users will not be affected by the removal until they make the effort to switch away from the deprecated apps.
- Remove the .env variables for the deprecated apps. Existing users who have these variables will not lose them, DS will just move them to the bottom of the file. The GUI will still pick them up as well. Echoing the above point, users who already have these apps in place should not be affected until they make the efforts to stop using the deprecated apps. New users will not be presented the option to use the deprecated apps unless they go out of their way to add the .env vars on their own.

- [x] Testing (mostly for the renamed unifi)

#### Learning

Deprecated images are listed here https://fleet.linuxserver.io/
We're only doing the LSIO ones for now as there's not a decent indication from most other image sources whether an image is deprecated or not.

## Requirements


nemchik:
I'm adding Samba to the list being deprecated. It's had problems since day 1 and they have not been addressed.

PR #709, opened 2019-05-21
Hostnames

### LINKED ISSUE #708: Only include hostname for compatible network modes
The `hostname` directive we have added to the compose templates is not allowed in certain network modes.

https://docs.docker.com/compose/compose-file/#domainname-hostname-ipc-mac_address-privileged-read_only-shm_size-stdin_open-tty-user-working_dir

https://docs.docker.com/engine/reference/run/#uts-settings---uts

The documentation only lists the `hostname` directive as being incompatible with `host` network mode, but I believe I recall a conversation with someone on Discord stating that they could not set the network mode of a container to use another container or service due to the `hostname` directive and had to do it through an override without the `hostname` directive in the compose. This should be investigated further and confirmed before changes are made.

nemchik:
## Purpose

This closes #708

## Approach

Reduce some of the output from `generate_yml` and split the `hostname` directive out from the main yml file for each app and into their own file that can be included as needed. For now it is included anytime the ports file is included. This is not combined with the ports file in case it could be included with other network modes in the future.

- [x] Testing
- [ ] After this merges, tag any other open PR involving adding new apps in the comments below requesting updates to reflect the new changes
- [x] Update contributing guideline

## Requirements

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -21,19 +21,20 @@ More important than being beautiful is being functional. This repository is prim
 - Should be formatted with [https://prettier.io/](https://prettier.io/)
 - Are separated into multiple files:
   - `<appname>.yml` is the main YAML template for an app and should have elements in the following order:
-    - `# APPNICENAME` comment on line 1, must match <appname> exactly but can have mixed case. Ex: Portainer vs PORTAINER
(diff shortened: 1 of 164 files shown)


nemchik:
Affected PRs that will probably need to update with the new method of including `hostname`:

#702
#647
#639
#638
#637
#634

Issue #676, opened 2019-05-03
Documentation for all apps


nemchik:
I think we should make a documentation page for every app in DS whether we have something special to say about it or not. Each page should have at least some basic instructions about where the app can expect to see files like `/movies` or how to connect it to other apps (ex: on the page for ombi explain to enter `sonarr` as the hostname to connect to sonarr and stuff like that).

Each app page should be included in the navigation and the filenames should be lowercase.

nemchik:
This is open to anyone to contribute, even one page at a time. Help is welcome.

nemchik:
I've started a branch to create placeholder pages for all apps that are currently missing pages.

nemchik:
#700 has created placeholder pages, but I'd like to keep this open until a decent majority of the pages have real content

nemchik:
Keep this

mjmaurer:
Are the docs open source?

nemchik:
Yes. https://github.com/GhostWriters/DockSTARTer.com

Also just visiting the pages of https://dockstarter.com/ you'll see edit buttons on the top right corner of every page. Changes are reviewed and approved before becoming live.

nemchik:
This will be ongoing and doesn't need an issue here.

PR #719, opened 2019-05-30
Update Organizr Documentation

angrycuban13:
Not sure if this is what you are looking for. Mostly copy/paste from Org's GitHub page. If you are looking for something a little bit more specific let me know :)

## Purpose

Addressing #676 

## Approach

Provided some basic documentation of what Organizr does.

#### Learning

_Describe the research stage_

_Links to blog posts, patterns, libraries or addons used to solve this problem_

## Requirements

### DIFF
--- a/docs/apps/organizr.md
+++ b/docs/apps/organizr.md
@@ -1,1 +1,36 @@
-Placeholder page
+# What is Organizr?
+
+Do you have quite a bit of services running on your computer or server? 
+
+Do you have a lot of bookmarks or have to memorize a bunch of ip's and ports? Well, Organizr is here to help with that. 
+
+Organizr allows you to setup "Tabs" that will be loaded all in one webpage. You can then work on your server with ease. 
+
+Want to give users access to some Tabs? No problem, just enable user support and have them make an account. 
+
+Want guests to be able to visit too? Enable Guest support for those tabs.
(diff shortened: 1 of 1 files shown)


nemchik:
I'm torn on this. It's great info, but I think for this (and many apps) it makes sense to provide a brief overview (a bit smaller than this) and link to the place where this documentation came from and have more specific information about how to see it up on the context of DS (which is kind of generic to docker environments). Things like what folder to set the database to save to, or how to setup tabs and what URLs to use and what to put for the homepage and sso settings for Plex and such. If we were talking about sonarr I'd think a good example would be what folder to set for the root tv folder.

On the other hand org can be configured MANY ways, and some may depend on other services like letsencrypt proxy being setup first. So I'm not sure if details about specific config make sense. Apps like sonarr will only have one place where you'll save TV, or one specific way to setup a connection to whatever download client, so those might be easier.

I'm open to suggestions, but I think we should try to cover what we are asked the most for help with. I think people can see a list of features on org's page :smile:

angrycuban13:
Yeah that's true. I can just throw a hyperlink pointing to Org. Let me think of what little tidbits I can add and just point them to our documentation as well :)

angrycuban13:
I'll close this PR and create a new one with some documentation.

PR #728, opened 2019-06-11
Update opencollective backers / sponsors

znarf:
This PR update the Readme to track 'backer' and 'sponsor' tiers on Open Collective, instead of relying on the confusing default behavior for GitHub readmes.

Current version: https://github.com/GhostWriters/DockSTARTer/blob/master/README.md
New version: https://github.com/znarf/DockSTARTer/blob/patch-1/README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,19 +99,19 @@ Support the project by donating on Open Collective.
 
 ### Backers
 
-[![Backers on Open Collective](https://img.shields.io/opencollective/backers/DockSTARTer.svg?style=flat-square&color=607D8B)](https://opencollective.com/DockSTARTer#backer)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -99,7 +99,7 @@ Support the project by donating on Open Collective.
 
 ### Backers
 
-[![Backers on Open Collective](https://img.shields.io/opencollective/tier/DockSTARTer/7408.svg?style=flat-square&color=607D8B)](https://opencollective.com/DockSTARTer#backer)
+[![Backers on Open Collective](https://img.shields.io/opencollective/tier/DockSTARTer/7408.svg?style=flat-square&color=607D8B&label=backers)](https://opencollective.com/DockSTARTer#backer)
```
znarf: I did that for the other badge because it was answering with "new tier" (problem on shield.io side). Not needed here but doesn't hurt if the goal is to be consistent.

nemchik: Right, I saw that. I figured consistency would be good.

PR #780, opened 2019-08-03
Start adding force flag

nemchik:
## Purpose
Certain installs can be skipped if the minimum version of the required app is already installed. This will allow forcing DS to install the latest even if the minimum is already installed.

## Approach
Adding a `-f`/`--force` flag that users can include in their CLI runs. Adjust the default behaviour of purging too account for the force flag. Some commands will now have prompts on CLI that previously did not, however these prompts can be skipped with the new force flag.

- [x] Testing `ds -u origin/force-flag`

## Requirements


nemchik:
This is still valid, just have not had time to finish it.

.scripts/appvars_purge.sh
```diff
@@ -17,9 +17,6 @@ appvars_purge() {
     fi
 
     local PREPROMPT=${PROMPT:-}
```
MagicalCodeMonkey: With the moving of the PROMPT check I think PREPROMPT is no longer needed

.scripts/appvars_purge_all.sh
```diff
@@ -5,9 +5,6 @@ IFS=$'\n\t'
 appvars_purge_all() {
     if grep -q '_ENABLED=false$' "${SCRIPTPATH}/compose/.env"; then
         local PREPROMPT=${PROMPT:-}
```
MagicalCodeMonkey: Same comment as https://github.com/GhostWriters/DockSTARTer/pull/780/files#r381522530. Can discuss there but just want to point this out for our future selves.

.scripts/appvars_purge.sh
```diff
@@ -17,9 +17,6 @@ appvars_purge() {
...
-    if [[ ${CI:-} != true ]] && [[ ${PROMPT:-} != "GUI" ]]; then
-        PROMPT="CLI"
-    fi
     if [[ ${CI:-} == true ]] || run_script 'question_prompt' "${PROMPT:-}" N "Would you like to purge these settings for ${APPNAME}?\\n\\n${APPVARS}"; then
```
MagicalCodeMonkey: I think this should this be changed to
`if [[ ${CI:-} == true ]] || [[ ${PROMPT:-} == "FORCED" ]] || run_script 'question_prompt' "${PROMPT:-}" N "Would you like to purge these settings for ${APPNAME}?`
so that they are always purged if PROMPT=FORCED

.scripts/appvars_purge_all.sh
```diff
@@ -5,9 +5,6 @@ IFS=$'\n\t'
...
-        if [[ ${CI:-} != true ]] && [[ ${PROMPT:-} != "GUI" ]]; then
-            PROMPT="CLI"
-        fi
         if [[ ${CI:-} == true ]] || run_script 'question_prompt' "${PROMPT:-}" N "Would you like to purge variables for all disabled apps?"; then
```
MagicalCodeMonkey: Same comment as  https://github.com/GhostWriters/DockSTARTer/pull/780/files#r381523972. Can discuss there but just want to point this out for our future selves.

main.sh
```diff
@@ -476,7 +485,9 @@ main() {
         exit
     fi
     # Run Menus
-    PROMPT="GUI"
+    if [[ ${PROMPT:-} != "FORCED" ]]; then
+        PROMPT="GUI"
+    fi
```
MagicalCodeMonkey: Thinking that this isn't actually right and that the PROMPT=FORCED should only apply to command-line args. Otherwise, question prompts for things like setting variables will be skipped when going through the GUI.

nemchik: Valid point. I believe you are correct.

main.sh
```diff
@@ -350,6 +356,9 @@ main() {
     fi
     # Repo Check
     local PROMPT
+    if [[ ${FORCE:-} == true ]]; then
+        PROMPT="FORCED"
```
nemchik: Can we use `PROMPT="FORCE"` (no `D`)?

.scripts/question_prompt.sh
```diff
@@ -7,6 +7,9 @@ question_prompt() {
     local DEFAULT=${2:-Y}
     local QUESTION=${3:-}
     local YN
+    if [[ ${PROMPT:-} != "FORCED" ]] && [[ ${PROMPT:-} != "GUI" ]]; then
+        PROMPT="CLI"
+    fi
```
nemchik: We can accomplish this by changing line 6 to `local PROMPT=${1:-CLI}`
Basically if it's not set then assume we need CLI, but if it is set then roll through the current logic. Wouldn't hurt to explicitly define how FORCE is handled by adding
```bash
        elif [[ ${PROMPT:-} == "FORCE" ]]; then
            YN=${DEFAULT}
```
above the else on line 36

.scripts/appvars_purge.sh
nemchik: Adding `|| [[ ${PROMPT:-} == "FORCED" ]]` would be redundant to the logic in the `question_prompt` script. This would negate the need to call the other function at all, but I don't think keeping it simple might (fewer added conditionals) would be beneficial since the functionality is already there to handle it.

.scripts/appvars_purge.sh
nemchik: I think you MIGHT be right on this, but I remember adding it here for a specific reason and of course can't remember. I'll check the git blame and see if that refreshes my memory!

.scripts/appvars_purge.sh
MagicalCodeMonkey: Should the default be changed to `Y` instead of `N` then? That way it will purge all the variables if PROMPT=FORCE. 
If so, `[[ ${CI:-} == true ]] ||` could then also be removed since it would be handled by `question_prompt`

.scripts/appvars_purge.sh
nemchik: Ok I've re-reviwed this PR and I think you're entirely correct that PREPROMPT is no longer needed anywhere anymore and can be removed from all files.

.scripts/appvars_purge.sh
nemchik: Yes to changing the default to `Y` and removing the CI condition.

.scripts/appvars_purge_all.sh
nemchik: Agreed

.scripts/appvars_purge_all.sh
nemchik: Agreed

nemchik:
@MagicalCodeMonkey I've resolved all the conversations we were having about adjustments. If you wouldn't mind having a look at the current state of this PR and adding a review (approve or request changes).

Issue #729, opened 2019-06-13
NGINX Proxy Manager


jodfie:
**Is your feature request related to a problem? Please describe.**
NGINX Proxy Manager is a great user friendly reverse proxy manager. 

**Describe the solution you'd like**
Integrate Jc21/Nginxproxymanager

**Additional context**
https://github.com/jc21/nginx-proxy-manager

nemchik:
Keep this.

I'm interested in this but haven't had time to look into it yet.

Bockiii:
Instead of using nginx, research traefik. It's the new and better way to do this and has a really good docker integration.

nemchik:
Better it's subjective.

Nginx has been around longer, has more features, is very well documented and has a large and active community of users.

dutran88:
Any updates with this?

nemchik:
It's not off the table, but I've been very busy with work and haven't had a chance to put it together. You're welcome to submit a PR to add it, or use overrides to run it personally until I've got time to look into it.

kazo0:
I'd like to have this very much. I can take a crack at implementing it if you don't mind.

nemchik:
Absolutely!

kazo0:
Sweet, you can assign it to me. I'll look into it this week/weekend!

Bockiii:
I not only take back my foolish words, but also added the proxy manager to dockstarter. I didn't test it (because I dont have a dev-environment right now), but it looks exactly like the additions of codeserver, monitorr and dasshio, so I assume it should work.

Bockiii:
Also, fyi @ op: I took the docker from jlesage, not jc21 because it includes the database and config, so its less setup.

jodfie:
**> Also, fyi @ op: I took the docker from jlesage, not jc21 because it includes the database and config, so its less setup.**
> 
> Only downside: no arm version.

i didn't know anyone else had created a version so cool!

andrewr78:
I know this is a bit stale - but any further word on integrating this into dockstarter?

nemchik:
I looked into this just a few weeks ago. The issue is it requires you make a config file ahead of time and bind mount the file into the container. DS doesn't provide any container that requires this kind of setup. If we include a file bind mount in the DS templates and the file doesn't exist ahead of starting the container compose will create a folder (not a file) which just leads to confusion.

You can still run this via overrides.

PR #808, opened 2019-10-08
#729 : Add nginx-proxy-manager

Bockiii:
## Purpose

Add nginx-proxy-manager

Not tested, but should work.


Bockiii:
Hi... i'm an idiot ^^

I saw that the jc21 docker is used as the base for the jlesage docker, but this one (jlesage) adds the database and config to the docker, so it's a one-stop-shop.

So a new pr with the same content :)

nemchik:
Let's use jc21/nginx-proxy-manager since it's the official source and has arm images.

Their example compose instructs to make a volume for a file. DS never does this as it's a very inconsistent experience for end users. Instead we need to make a volume for the folder. It seems they also have multiple volumes that I think would be just fine to be part of the config volume. So config/data and whatnot.

Bockiii:
Hi,

I am creating the volume for the folder (see the main yml) as that's enough, no need to reference the files. So yes, the one folder /config is enough.

The main problem with the jc21 image is, that it needs an additional mariadb container that needs to be setup and linked. The jlesage container bases off the jc21 and adds the mariadb to the container.

I already put in a feature request to jlesage to maybe get an arm image, but I think it's more effort to use the jc21 image than this.

nemchik:
As much as I appreciate the included DB I'm going to say the arm images win and I'd rather use the jc21 images.

There's a few other apps in DS that require a database, so I'm not too worried about that.

Bockiii:
Let me check how that works and I'll redo the pr

timelf123:
@Bockiii have you redone with the other image? if not I might take a stab at it, let me know

Bockiii:
Hi, no tbh. I use the jc21 container and like it a whole lot more than the upstream container so I didn't bother to change it.

If you want to, go for it

xivor:
I have successfully set up the JC21 image yesterday, with the separate mariadb image.  I have it configured as an override right now with a configuration very similar to the one proposed here.  I prefer jc21 over jselage, and it is regularly updated still. My only wish was that the DB was included on the image rather than running two images, but since I am not trying to host a new docker image, I am not going to complain much about it.

I have never set up LE as fast as I did last night, once I got the image properly integrated.  I believe this app is essential for streamlining the most painful part of setup of a homelab: reverse proxies management.

nemchik:
@xivor feel free to post your override here and i might give some feedback, or open a PR to add the image to DS and i'll review it.

Bockiii:
> I prefer jc21 over jselage, and it is regularly updated still.

Just a quick note on that: jc21 is doing a bunch of stuff with arm and configuration topics, but the actual software is the same version on both containers (2.0.14). So yes, while it shows more activity, both containers are on the same software version.

xivor:
> > I prefer jc21 over jselage, and it is regularly updated still.
> 
> Just a quick note on that: jc21 is doing a bunch of stuff with arm and configuration topics, but the actual software is the same version on both containers (2.0.14). So yes, while it shows more activity, both containers are on the same software version.

I did notice that, actually. In the end, Jselage is merely rebuilding a new image with the same original base images that jc21 uses. Effectually, both options are suitable. Given that we already have mariadb integrated with DS, and it already has use-cases with other applications in DS, I am leading more towards keeping with traditional dual-image. Or maybe we can convince linuxserver to put together an image and maintain it. Then its standardized with most of the DS stack.

I am setting up a repository for this override.  Give me some time to get that together today, and I will post the link here. I am somewhat new to contributing on Github, so bear with me.

nemchik:
For the sake of reference i'm going to link to both of these projects (probably should have sooner in the comments):

https://github.com/jc21/nginx-proxy-manager
https://github.com/jlesage/docker-nginx-proxy-manager

The jlesage repo just packages the app from the jc21 repo along with a built in database from what I can tell. This is fine and good, and I don't dislike the implementation, but it only offers x86_64 images while jc21 offers ARM images as well as x86_64. DS never uses different images on different architectures for the same app (in case of differences in the images), so we'd want to pick only one of these two projects. It's not always the most official image that gets picked, it's the one that suits DS the best. For example apache's guacamole is not very easy for end users to setup, so we use oznu's image (which has ARM images and includes a built in database vs apache's official image which does not). Plex's official image does not have ARM support, so we use LSIO's that does.

To be perfectly clear about the goal of this project; it's to help people get started using docker. We want DS to be able usable on as many systems and by as many people as possible, which means including ARM whenever possible (raspberry pi's are very popular) and choosing images that are easy to setup and use whenever possible.

In this case I'm standing by jc21's image being the winner of the two for our needs.

xivor:
I have updated a repository with the override and any setup criteria. I think we should add one additional item to the COMPOSE file, however.  There is a step where we have to copy a config file to the config directory or it will not work.  I think we can add a COMMAND: bash -c "cp config.json ~/.config/appdata/proxyserver". Any other thoughts on a better way, feel free to add.

https://github.com/xivor/dockSTARTer-overrides

xivor:
I discussed a bit with Nemchik on Discord and because this application requires an external config file, it does not fit the ease-of-use principle that DS consistently follows.  That being said, I took the initiative to create a Feature Request issue for JC21's image for an all inclusive image that contains the config file.  I also recommended docker environment variables for the config file values, so they can still be modified by the user through compose/portainer, etc.

https://github.com/jc21/nginx-proxy-manager/issues/272

Bockiii:
since you're going to start a new one anyway, i will close this pr

Issue #847, opened 2020-01-10
samba - volume name is too short


GMFalka:
**Describe the bug**
With full setup option, dockstarter cannot compose samba, because:

`ERROR: for samba  Cannot create container for service samba: create .: volume name is too short, names should be at least two alphanumeric characters
`

**System**

Clean install Ubuntu 18.04

Any help would be appreciated!

nemchik:
Can you run `grep SAMBA_ ~/.docker/compose/.env ` and post the output here? (sanitize any passwords, but let us know it's filled out)

GMFalka:
> Can you run `grep SAMBA_ ~/.docker/compose/.env ` and post the output here? (sanitize any passwords, but let us know it's filled out)

Thanks for the quick reply.
Its kinda late here, can i get back to you next week or after that?
thanks again

markiljas:
I have the same issue, so posting my  `grep SAMBA_ ~/.docker/compose/.env`. I'm also on a fresh Ubuntu 18.04 install.

```
SAMBA_ENABLED=false
SAMBA_NETWORK_MODE=
SAMBA_NMBD=true
SAMBA_PASSWORD=*mypasswordhere*
SAMBA_PORT_137=137
SAMBA_PORT_138=138
SAMBA_PORT_139=139
SAMBA_PORT_445=445
SAMBA_SHARENAME=Alfred
SAMBA_USERNAME=hp
SAMBA_WORKGROUP=WORKGROUP
```

Thanks!

PabloHuerfano:
Im am relatively sure the issue is due to this line in the samba.yml file provided in dockstarter. I would advise removing it from the samba.yml file. 

```
    volumes:
      - ./:/${SAMBA_SHARENAME}/compose
```

Along, the same issue why are volumes predefined in the dockstarter samba.yml. There is not an effective way to remove the predefined volumes and allow a user to choose which folder the samba share will have. This has led me to stop using the Samba option in dockstarter and just spinning one up with the compose override file.

nemchik:
Sorry for the hold up here. I believe @PabloHuerfano is correct and has opened a https://github.com/GhostWriters/DockSTARTer/pull/853 to resolve this issue.

@GMFalka if you want to test and confirm this fix please run `ds -u PabloHuerfano/master` and then `ds -c`

PR #853, opened 2020-01-16
Fixing samba.yml

PabloHuerfano:
Solves an issue with "samba - volume name is too short" error at composing

**Purpose**
DockStarter builtin Samba option wasn't starting

**Approach**
Removed line causing the issue in samba.yml
This leads to lose of compose files being exposed through Samba share.

Why are these specific volumes predefined?
No easy way for the user to adjust them through dockstarter.


nemchik:
This looks good to me. I'd like a confirmation from @GMFalka that testing `ds -u PabloHuerfano/master` resolves the issue.

This closes https://github.com/GhostWriters/DockSTARTer/issues/847

nemchik:
Even prior to these changes I did not see the error reported in #847 and these changes look good to me so I'm going to merge them. If #847 is not resolved we can reopen it and continue looking into the issue.

nemchik:
Found another related issue. Merging this and another fix shortly.

PR #858, opened 2020-01-18
Add bitwarden vars

nemchik:
**Purpose**

**Approach**

**Learning**
https://github.com/dani-garcia/bitwarden_rs/blob/master/.env.template

**Requirements**


rob151515:
I'm having a problem with new Bitwarden containers now where if I do not specify something for BITWARDEN_SMTP_FROM & BITWARDEN_SMTP_HOST in my env's, the container will error and won't start up properly. It looks like there is no way to now specify an empty value for these variables? (I don't want to put anything in these variable as I don't won't email verification).

nemchik:
@rob151515 try putting `localhost` for the host and `fake@fake.com` for the from?

rob151515:
Thanks @nemchik I had already added fake values for the variables to get the container up and running but was getting a message when logged into Bitwarden about email verification which I had assumed was because we are now setting BITWARDEN_SMTP_FROM & BITWARDEN_SMTP_HOST. Oh well, at least it's working. Cheers.

PR #871, opened 2020-02-10
Add API Token option to CloudflareDDNS

MagicalCodeMonkey:
**Purpose**
Cloudflare has added API Tokens for more granular control of access to your account. The API Key can still be used but API Tokens are a more secure approach.

**Approach**
Adding the supported variable, CF_APITOKEN, to the yml makes this possible.

**Learning**
After trying to use an API Token for the API Key, the container provided an error that the configuration was invalid. Looking at the documentation at https://hub.docker.com/r/hotio/cloudflare-ddns/ the container supports the use of API Tokens in lieu of using an API Key (CF_APIKEY) with CF_USER

**Requirements**

### DIFF
--- a/compose/.apps/cloudflareddns/cloudflareddns.labels.yml
+++ b/compose/.apps/cloudflareddns/cloudflareddns.labels.yml
@@ -3,6 +3,7 @@ services:
     labels:
       com.dockstarter.appinfo.description: "Free service which will point a DNS to an IP of your choice"
       com.dockstarter.appinfo.nicename: "CloudFlareDDNS"
+      com.dockstarter.appvars.cloudflareddns_apitoken: ""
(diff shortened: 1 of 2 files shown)


nemchik:
Will adding this cause issues for existing users if they don't set a token (leave it blank) but continue to use an API key?

MagicalCodeMonkey:
Doesn't appear to have any issues if CLOUDFLAREDDNS_APITOKEN isn't set, but CLOUDFLAREDDNS_APIKEY and CLOUDFLAREDDNS_APIUSER are, then it works. 

I also tried having all three set and it appears to take the CLOUDFLAREDDNS_APITOKEN and ignores CLOUDFLAREDDNS_APIKEY & CLOUDFLAREDDNS_APIUSER.

PR #872, opened 2020-02-13
Docker overrides

nemchik:
Making this for the purpose of commenting/review


.scripts/docker_overrides_compile.sh
```diff
@@ -0,0 +1,84 @@
...
+            else
+                warn "${path//${DOCKER_OVERRIDES_DIR}\//} is not valid yml and has not been included. Please check your syntax."
+                VALIDATION_ERRORS=$((VALIDATION_ERRORS + 1))
+            fi
+        fi
+    done < <(find ${DOCKER_OVERRIDES_DIR}/* -type f -not -name "original_overrides.yml" -prune)
+    shopt -u dotglob
+
+    # Include the user's original overrides, if they existed
+    if [[ -f "${DOCKER_OVERRIDES_DIR}/original_overrides.yml" ]]; then
```
nemchik: User defined file was moved here from line 14, and then the user would need to know that they should always be able to edit this file and should never edit the docker-compose.override.yml that's pieced together by this script. The comments added from lines 32-37 indicate to not edit the docker-compose.override.yml file but don't tell the user where their original file ends up.

.scripts/docker_overrides_compile.sh
```diff
@@ -0,0 +1,87 @@
...
+    local DOCKER_OVERRIDES_DIR
+    DOCKER_OVERRIDES_DIR=$(run_script 'env_get' DOCKEROVERRIDESDIR)
+    local VALIDATION_ERRORS=0
+
+    info "Running Docker Overrides Compiler"
+    # Move the user's existing docker-compose.override.yml, if it exists
+    if [[ -f "${SCRIPTPATH}/compose/docker-compose.override.yml" ]]; then
+        if ! grep -q '# GENERATED BY DOCKSTARTER' "${SCRIPTPATH}/compose/docker-compose.override.yml"; then
+            warn "Found user-created docker-compose.overrides.yml file. Moving to ${DOCKER_OVERRIDES_DIR}/original_overrides.yml"
+            mv "${SCRIPTPATH}/compose/docker-compose.override.yml" "${DOCKER_OVERRIDES_DIR}/original_overrides.yml"
```
nemchik: Add `| info "didn't work"` or something (using the appropriate logging level and message) at the end

.scripts/docker_overrides_compile.sh
```diff
@@ -0,0 +1,87 @@
...
+        fi
+    fi
+    # Check for the directory and files should this be used from the command line
+    if [[ ! -d ${DOCKER_OVERRIDES_DIR} ]]; then
+        error "${DOCKER_OVERRIDES_DIR}/ does not exist. Create it and populate it with files to validate."
+        return
```
nemchik: Are you doing a return here because you intend for the script to end? Or you want to catch this scenario by whatever script is calling this one?

.scripts/docker_overrides_compile.sh
```diff
@@ -0,0 +1,87 @@
...
+        echo "echo '################################################################' > ${SCRIPTPATH}/compose/docker-compose.override.yml"
+        echo "echo '#                   GENERATED BY DOCKSTARTER                   #' >> ${SCRIPTPATH}/compose/docker-compose.override.yml"
+        echo "echo '#                        DO NOT MODIFY!                        #' >> ${SCRIPTPATH}/compose/docker-compose.override.yml"
+        echo "echo '# If you already had this file, it has been retained and moved #' >> ${SCRIPTPATH}/compose/docker-compose.override.yml"
+        echo "echo '# See DockSTARTer documentation for more information           #' >> ${SCRIPTPATH}/compose/docker-compose.override.yml"
+        echo "echo '# https://dockstarter.com/advanced/overrides/                  #' >> ${SCRIPTPATH}/compose/docker-compose.override.yml"
+        echo "echo '################################################################' >> ${SCRIPTPATH}/compose/docker-compose.override.yml"
+        # Merge any included yml
+        echo '/usr/local/bin/yq-go m '\\
+        echo "${SCRIPTPATH}/compose/.reqs/*.yml \\"
```
nemchik: Consider using this method for multi-line redirection https://github.com/GhostWriters/DockSTARTer/blob/41c41911d03dde8c49786d42fd6b79faead629bb/.scripts/yml_merge.sh#L12-L16

.scripts/docker_overrides_compile.sh
MagicalCodeMonkey: Debating between
`mv [...] || warn "Unable to move file. A backup can be found at ${DOCKERCONFDIR}/.compose.backups/"`
and just
`mv [...] || fatal ""Unable to move file."`

Thoughts?

.scripts/docker_overrides_compile.sh
MagicalCodeMonkey: End the script. Would `exit` be more appropriate?

.scripts/docker_overrides_compile.sh
MagicalCodeMonkey: Isn't that what is happening on lines 32 - 44, or do you mean changing `echo "${SCRIPTPATH}/compose/.reqs/*.yml \\"` to not be a wildcard and use the v1 and v2 lines instead?

.scripts/docker_overrides_compile.sh
nemchik: Use fatal on the line above instead of error

.scripts/docker_overrides_compile.sh
nemchik: Wait I'm reading this wrong. You're doing it for the runfile, but I was looking at 34-40 all going to the same place.

.scripts/docker_overrides_compile.sh
nemchik: Will there be loss of user data if you do not fatal? Will the script still work from there?

MagicalCodeMonkey: No data loss since the original file would be backed up, but it wouldn't be included in the generation until the user copied the backup into the overrides directory. So, I'm thinking a `fatal` would be the better option overall.

.scripts/docker_overrides_compile.sh
MagicalCodeMonkey: Oh, got it! I'll take a look into how I can put those header comments into the runfile so that they are put into the generated yml without doing output redirect on each line but just the final one.

.scripts/docker_overrides_compile.sh
MagicalCodeMonkey: Made it `fatal`, just to be safe.

MagicalCodeMonkey:
Closing this and moving everything to DSAC - https://github.com/GhostWriters/DSAC/pull/28

PR #896, opened 2020-04-22
[ADD] Deemix

Bockiii:
**Purpose**
Added "Deemix". Application for downloading music from Deezer. Successor to deezloader rmx which was discontinued.

**Approach**
Website with login and search functionality

-

- [x ] Use github checklists. When solved, check the box and explain the answer.

**Learning**
-

**Requirements**
-

- [x ] These changes meet the standards for [contributing](https://github.com/GhostWriters/DockSTARTer/blob/master/.github/CONTRIBUTING.md).
- [x ] I have read the [code of conduct](https://github.com/GhostWriters/DockSTARTer/blob/master/.github/CODE_OF_CONDUCT.md).

### DIFF
--- a/compose/.apps/deemix/deemix.yml
+++ b/compose/.apps/deemix/deemix.yml
@@ -1,0 +1,17 @@
+services:
+  deemix:
+    container_name: deemix
+    environment:
+      - PGID=${PGID}
+      - PUID=${PUID}
+      - TZ=${TZ}
+    logging:
+      driver: json-file
+      options:
+        max-file: ${DOCKERLOGGING_MAXFILE}
+        max-size: ${DOCKERLOGGING_MAXSIZE}
+    restart: unless-stopped
+    volumes:
+      - /etc/localtime:/etc/localtime:ro
+      - ${DOCKERCONFDIR}/deemix:/config
+      - ${MEDIADIR_MUSIC}:/downloads
--- a/compose/.apps/deemix/deemix.aarch64.yml
+++ b/compose/.apps/deemix/deemix.aarch64.yml
@@ -1,0 +1,3 @@
+services:
+  deemix:
+    image: bocki/deemix
--- a/compose/.apps/deemix/deemix.armv7l.yml
+++ b/compose/.apps/deemix/deemix.armv7l.yml
@@ -1,0 +1,3 @@
+services:
+  deemix:
+    image: bocki/deemix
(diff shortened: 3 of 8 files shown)


compose/.apps/deemix/deemix.yml
```diff
@@ -0,0 +1,17 @@
...
+    logging:
+      driver: json-file
+      options:
+        max-file: ${DOCKERLOGGING_MAXFILE}
+        max-size: ${DOCKERLOGGING_MAXSIZE}
+    restart: unless-stopped
+    volumes:
+      - /etc/localtime:/etc/localtime:ro
+      - ${DOCKERCONFDIR}/deemix:/config
+      - ${MEDIADIR_MUSIC}:/downloads
```
nemchik, line 17: This should be:
```yaml
- ${DOWNLOADSDIR}:/downloads
- ${MEDIADIR_MUSIC}:/music
```

Bockiii: Deemix directly downloads music and does folder/file and tag naming. The usecase should be do directly download into your music folder. downloading to an intermediate folder does not provide any benefits. 

Also, there is no /music folder in the container, it downloads to /downloads.

So if you don't want it to download directly to the music folder, I would just switch that mapping from mediadir_music to downloadsdir. I would still advise against it because that would need further automation to move files afterwards.

Your choice.

Bockiii: @nemchik can you tell me if you still want the changes after my info?

nemchik: Can the app (in the app settings) be configured to use `/music` ?

Really I would just want to keep things consistent with the rest of how DS sets up volumes.

Bockiii: Not without additional work on the dockerfile and init scripts (because of owner priveledges and so on).

We can switch it to downloadsdir:/downloads and leave the /music. That would mean that the user always needs to manually move the downloaded files into his music folder or install further automation.

k. I'll do that.

Bockiii:
will open a new

compose/.apps/deemix/deemix.yml
nemchik: I know it means extra work for the user, but I feel like keeping it consistent is important. Unless another directory entirely could be used? Ex: apps like nextcloud having a directory mounted.

PR #918, opened 2020-05-27
LSIO compose

nemchik:
**Purpose**
Use `linuxserver/docker-compose` to run `docker-compose` inside a docker container as needed instead of installing it natively using `python`. This also allows us to remove the use of python across the whole project. Docker officially offers this method, but only with x86_64. We previously attempted to use this method with Docker's official image but needed to support ARM CPU systems and settled on `python` and later improved `python3`. LSIO offering compose in a container and including multiarch builds resolves the need. If in the future Docker officially offers multiarch compose images we can easily switch to their official image and script.

**Learning**
https://github.com/linuxserver/docker-docker-compose
https://github.com/docker/compose/issues/6831

**Requirements**


idolize:
Cool!

nemchik:
It will affect new installs and anyone who updates DS. You can update via the menu (`ds`) or command `ds -u`.

The main benefit to existing users is not having the python dependencies. DS isn't uninstalling the python dependencies since other applications might need them, but nothing in DS will be making use of them anymore.

PR #988, opened 2020-07-22
Speedtest docs

angrycuban13:
# Pull request

**Purpose**

**Approach**

**Learning**
Describe the research stage

**Requirements**


docs/apps/speedtest.md
```diff
@@ -0,0 +1,25 @@
...
+
+### Notifications
+
+This application supports notifications to some of the most popular services such as Telegram and Discord. For Telegram, the `Bot Token` and `Telegram Chat ID` can be filled in using the Web GUI. However, for the Slack/Discord webhook it will need to be filled in using an environment variable on an [override](https://dockstarter.com/overrides/introduction).
+
+#### Resources
+
+##### [GitHub](https://github.com/henrywhitaker3/Speedtest-Tracker)
+
+##### [Docker Hub](https://hub.docker.com/r/henrywhitaker3/speedtest-tracker/tags)
```
nemchik, line 25: I think the hub links for most other apps are at the top and formatted a certain way. I'm not saying that's how we have to do it, but we should try to keep all the app pages the same. If we want to change the format (maybe use badges) we should do it across all at once in a separate PR.

angrycuban13: Ooooh I like the idea of badges actually. 

I was trying to think of a way to standardize how we link the upstream projects so badges would be awesome under the H1.

angrycuban13: Will use fe68992 to standardize the rest of our docs

PR #1023, opened 2020-07-31
couldn't connect to bitwarden

cerdelyan:
Port should be 80
https://github.com/GhostWriters/DockSTARTer/pull/636

# Pull request

**Purpose**

**Approach**

**Learning**
Describe the research stage

**Requirements**

### DIFF
--- a/compose/.apps/bitwarden/bitwarden.labels.yml
+++ b/compose/.apps/bitwarden/bitwarden.labels.yml
@@ -12,7 +12,7 @@ services:
       com.dockstarter.appvars.bitwarden_enabled: "false"
       com.dockstarter.appvars.bitwarden_invitations_allowed: "false"
       com.dockstarter.appvars.bitwarden_network_mode: ""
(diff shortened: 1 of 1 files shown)


nemchik:
DS doesn't allow apps to use well known ports https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers#Well-known_ports unless the app will not function without the port (ex pihole needs port 53, samba needs 445). Bitwarden can be accessed on http://your.ip:8013/ with the configuration DS comes with out of the box. Your better bet is to use a reverse proxy such as the letsencrypt container included in DS and then you can have https://bitwarden.you.com/

cerdelyan:
Ah my bad. I saw it mentioned to switch back to port 80 and the default bitwarden proxy_conf had port 80.

I wasn't able to connect using hostip:8013 until I changed it to 80 and hostip + https://bitwarden.example.com now work.

Not sure what was happening with me.

Issue #1056, opened 2020-08-25
PiHole


fordsteve:
# Bug report

**Describe the bug**
Couldnt install PiHole because ./docker-compose.yml is invalid because services.pihole.ports contains an invalid type, it should be a number or an object.

**To Reproduce**
After a clean install using DBTech's DockSTARTer guide on RPi4, Pihole created errors during install and broke the install chain. After deselecting PiHole in the apps install the automated setup could continue and i have containers etc installed in portainer. I went back to install PiHole via DS and the errors still appeared during install.

Can you please help?

Thanks

- Go to '...'
- Click on '....'
- Scroll down to '....'
- See error

**Expected behavior**

**Screenshots**

**System**
(please complete the following information):

- OS: [e.g. Debian]

**Additional context**

dylanmtaylor:
Hi, this issue is caused by the PIHOLE_SERVER_IP variable not being set. It defaults to 192.168.x.x, which is not valid. Configure it to be the IP address of your machine and it will work.

dylanmtaylor:
Maybe this could be changed? It does need to be set to work correctly https://github.com/GhostWriters/DockSTARTer/blob/master/compose/.apps/pihole/pihole.labels.yml#L13

fordsteve:
Thank you!

fordsteve:
I have corrected the IP addresses, but every time i run compose my changes get reset. How do i finish the setup process please?

dylanmtaylor:
> I have corrected the IP addresses, but every time i run compose my changes get reset. How do i finish the setup process please?

Set the IP via the menu in DockSTARTer. There is a menu item to set application variables

fordsteve:
thank you, i will update ds now.

fordsteve:
Thats now installed PiHole after i changed the IP to my dockstarter ip address. I will leave it to do any updates for a while as Portainer is telling me the container is unhealthy for some reason. i only changed the one IP in the menu?

I appreciate your time all, thanks for your support!

fordsteve:
I have just inspected the container in Portainer and there isn’t an IP address as shown below:

],
            "80/tcp": [
                {
                    "HostIp": "",
                    "HostPort": "8008"
                }
            ]

My IP has been set for the other ports though?

fordsteve:
Here is my port configuration for the container.


Port configuration	
0.0.0.0:446  443/tcp 
192.168.0.25:53  53/tcp 
[...]

PR #1057, opened 2020-08-25
Change default PiHole Server IP to 0.0.0.0

dylanmtaylor:
This is a possible solution for #1056. If the user does not set the server IP address they will currently get a confusing error and it will not allow docker-compose to run. This should at least get the container to build correctly. A message indicating that this should be set would also be greatly beneficial to users.


dylanmtaylor:
Testing done:

* I set the PIHOLE_SERVER_IP variable to 0.0.0.0 and re-ran compose, which worked successfully
* Using the new configuration, I was able to navigate to localhost:8008/admin and access PiHole's web interface

I believe this is a better default than compose not working at all.

dylanmtaylor:
@nemchik This one is ready for review as well

nemchik:
This feels like a good change, but the reason we don't have it default to this already is it still can be a conflict, often with resolv listening on 53, but this could work out of the box for some. All of the documentation on our pihole page is out of date. The current setup (before this change) doesn't require any of the current instruction on the current pihole documentation page. Would you be willing to update that page and replace everything with instructions relevant to this change? We would want to include the idea that you can set the IP to your server IP address (ex `192.168.1.5` or whatever) in case `0.0.0.0` doesn't work out of the box. Basically there should be no need to modify the host system, just the docker variables.

dylanmtaylor:
> Would you be willing to update that page and replace everything with instructions relevant to this change?

Sure, let me take a look at the docs and see what needs changed.

dylanmtaylor:
@nemchik - I made a commit with new documentation, what do you think of that?

docs/apps/pihole.md
```diff
@@ -5,52 +5,8 @@
...
-
-### Name resolution for localhost
-
-In most cases it might be required to set your localhost name in `/etc/hosts`
-
-```hosts
-127.0.0.1    machinename.localhost    machinename
-127.0.0.1    domain.com
-```
+By default, Pi-hole is configured to bind to `0.0.0.0` on port 53. This configuration should work out of the box for many people, however, if you encounter issues with this, you may need to disable or reconfigure any conflicting services running on port 53, such as `systemd-resolved` or set the `PIHOLE_SERVER_IP` variable to your server's IP address (e.g. 192.168.1.5). To stop systemd-resolved from listening on port 53, you can set `DNSStubListener=no` in `/etc/systemd/resolved.conf`.
```
nemchik, line 12: I love how minimal this is, but I would take it a step further and recommend the `PIHOLE_SERVER_IP` option first and then list the option to modify the resolved.conf as an alternative that may be needed for advanced configurations.

nemchik:
By the way are you on our discord?

dylanmtaylor:
> By the way are you on our discord?

Yes, as Dylan#9414

PR #1060, opened 2020-08-25
Don't only check if /usr/bin/docker-compose is a link

dylanmtaylor:
On Arch Linux, /usr/bin/docker-compose is a script not a symlink, so this fails. Changing from -L which checks if it is a symlink to -e which checks if it exists fixes the problem

### DIFF
--- a/.scripts/install_compose.sh
+++ b/.scripts/install_compose.sh
@@ -28,7 +28,7 @@ install_compose() {
             # https://github.com/linuxserver/docker-docker-compose/blob/master/README.md#recommended-method
             info "Installing latest docker-compose."
             curl -fsL "https://raw.githubusercontent.com/linuxserver/docker-docker-compose/master/run.sh" -o /usr/local/bin/docker-compose > /dev/null 2>&1 || fatal "Failed to install docker-compose."
-            if [[ ! -L "/usr/bin/docker-compose" ]]; then
+            if [[ ! -e "/usr/bin/docker-compose" ]]; then
                 ln -s /usr/local/bin/docker-compose /usr/bin/docker-compose || fatal "Failed to create /usr/bin/docker-compose symlink."
             fi
             chmod +x /usr/local/bin/docker-compose > /dev/null 2>&1 || true


dylanmtaylor:
Even though the docker-compose package is installed, this fails as it's a regular file:

```
➜  ~ lsb_release -a; stat /usr/bin/docker-compose 
LSB Version:	1.4
Distributor ID:	Arch
Description:	Arch Linux
Release:	rolling
Codename:	n/a
  File: /usr/bin/docker-compose
  Size: 995       	Blocks: 8          IO Block: 4096   regular file
Device: fe02h/65026d	Inode: 42252181    Links: 1
Access: (0755/-rwxr-xr-x)  Uid: (    0/    root)   Gid: (    0/    root)
Access: 2020-08-25 16:30:48.043450590 -0400
Modify: 2020-07-02 15:44:27.000000000 -0400
Change: 2020-08-24 15:53:53.967170000 -0400
 Birth: 2020-08-24 15:53:53.967170000 -0400
```

nemchik:
My concern/question here is what if an actual binary or script exists in /use/bin and /use/local/bin ? Could result in a bit of a mix up.

Secondly Arch isn't supported by the docker-ce install script from get.docker.com so DS doesn't make efforts to support Arch.

dylanmtaylor:
> Secondly Arch isn't supported by the docker-ce install script from get.docker.com so DS doesn't make efforts to support Arch.

That's fine, but it works with Arch, so I think it makes sense to avoid breakage on the platform even if it's not officially supported. Plenty of Arch-based distros can run Docker just fine.

> My concern/question here is what if an actual binary or script exists in /use/bin and /use/local/bin ? Could result in a bit of a mix up.

I'm not sure exactly what is being asked. The line in the code is 
```
            if [[ ! -e "/usr/bin/docker-compose" ]]; then
```

Which means that creating the symlink will only be attempted if  /usr/bin/docker-compose does not exist. Are you referring to a binary possibly existing in another path? Changing this line doesn't affect that.

dylanmtaylor:
Maybe that line could be changed to check if the docker-compose command is in the path already, and if it isn't, create a symlink.

Something like this?
```
if ! $(command -v docker-compose); then
  ln -s /usr/local/bin/docker-compose /usr/bin/docker-compose || fatal "Failed to create /usr/bin/docker-compose symlink."
fi 
```

nemchik:
> Are you referring to a binary possibly existing in another path? Changing this line doesn't affect that.

Yes. If you already have a binary or script in both `/usr/bin/docker-compose` and `/usr/local/bin/docker-compose` and they are different for whatever reason (various other install methods) the idea is that we want DS to ensure there is only one actual file and the other should be a symlink because various distros expect it in one or the other location to be in `$PATH` so we try to cover both. Really if you (on Arch or any other distro) remove the file from both locations DS will set things up in a way that works.

dylanmtaylor:
Oh, I think I see what you're saying - we're downloading compose to /usr/local/bin and if that check fails, not making a symlink, so there might be two different versions installed leading to confusing behavior. Hmm. I think the issue might be with this line using an absolute path
```
        INSTALLED_COMPOSE=$( (/usr/local/bin/docker-compose --version 2> /dev/null || echo "0") | sed -E 's/.* version ([^,]*)(, build .*)?/\1/')
```

Since this is different to where the AUR docker-compose is installed to on Arch, it fails to detect it. I'm not actually sure what a good solution to this is. I'll have to think about it some more.

nemchik:
> Maybe that script could be changed to check if the docker-compose command is in the path already, and if it isn't, install it.

This script already checks for an existing docker compose and skips there install if you've already got a version above the minimum if you do the advanced install instructions from the readme.

nemchik:
We're calling the full path just out of best practice. It could be reasonable to symlink FROM `/usr/bin/docker-compose` TO `/usr/local/bin/docker-compose` but the issue is we'll have to come up with a lot of checks and balances. Anyway if you setup that symlink manually before running DS it should work.

PR #1061, opened 2020-08-26
Add 'Element' Matrix Web Server (Formerly Called Riot.im)

dylanmtaylor:
# Pull request

**Purpose**
Element (formerly Riot.im) is a Matrix web client.

**Approach**
Added new application

**Requirements**

### DIFF
--- a/compose/.apps/element/element.yml
+++ b/compose/.apps/element/element.yml
@@ -1,0 +1,17 @@
+services:
+  element:
+    container_name: element
+    environment:
+      - PGID=${PGID}
+      - PUID=${PUID}
+      - TZ=${TZ}
+    logging:
+      driver: json-file
+      options:
+        max-file: ${DOCKERLOGGING_MAXFILE}
+        max-size: ${DOCKERLOGGING_MAXSIZE}
+    restart: unless-stopped
+    volumes:
+      - /etc/localtime:/etc/localtime:ro
+      - ${DOCKERCONFDIR}/element:/app
+      - ${DOCKERSHAREDDIR}:/shared
--- a/compose/.apps/element/element.hostname.yml
+++ b/compose/.apps/element/element.hostname.yml
@@ -1,0 +1,3 @@
+services:
+  element:
+    hostname: ${DOCKERHOSTNAME}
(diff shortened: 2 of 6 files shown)


compose/.apps/element/element.yml
```diff
@@ -0,0 +1,17 @@
...
+      - TZ=${TZ}
+    logging:
+      driver: json-file
+      options:
+        max-file: ${DOCKERLOGGING_MAXFILE}
+        max-size: ${DOCKERLOGGING_MAXSIZE}
+    restart: unless-stopped
+    volumes:
+      - /etc/localtime:/etc/localtime:ro
+      - ${DOCKERCONFDIR}/element:/app
```
nemchik, line 16: Does the container put the whole app here?

If you run the container without this volume (meaning it will be missing the config.json file) is there anything else inside the container in `/app` ?

If yes to the above, if you run the container with this volume do the `/app` files get properly saved to `${DOCKERCONFDIR}/element` ?

dylanmtaylor: This is copied and pasted from another app, sorry, this change is still work in progress - I actually think this is wrong as element is not yet functional, that's why I have it marked draft.

dylanmtaylor: So it doesn't look like a default config.json is getting written to that conf dir. Looking back at this, I think I took the /app from the suggested binding
/etc/element-web/config.json:/app/config.json on the https://github.com/vector-im/element-web page

nemchik: What you might run into with this (i have seen this in the past with other apps) is the docker image actually has files in `/app/` and if you mount that as a volume those files would then not be present. This is often why you'll see containers recommend mounting an individual file instead of a whole directory. Unfortunately if we do this (volume `${DOCKERCONFDIR}/element/config.json:/app/config.json`) the user HAS to create the file needed (`${DOCKERCONFDIR}/element/config.json` including creating the parent folder) prior to starting the container for the first time otherwise the user will end up with a folder (`${DOCKERCONFDIR}/element/config.json/`) which is confusing as hell for users. Often this folder will also end up being owned by root so the user has to use sudo to delete it.

This may not be the case here, but this scenario is why I'm asking these questions (again, I've seen it with other images).

PR #1083, opened 2020-09-03
Make sure that /config/qBittorrent/downloads/ is mapped to the downloads directory

dylanmtaylor:


nemchik:
This is an interesting approach. I don't see any harm in it, but it might cause some confusion (although maybe less than NOT having it). I wonder if there are other containers that would benefit from the same change.

dylanmtaylor:
> This is an interesting approach. I don't see any harm in it, but it might cause some confusion (although maybe less than NOT having it). I wonder if there are other containers that would benefit from the same change.

I was very confused at first as to where my downloads were stored, as I set my downloadsdir to /mnt/downloads and then files were not there when testing out this container , and I imagine most users wouldn't be technical enough to figure it out easily.

dylanmtaylor:
> I wonder if there are other containers that would benefit from the same change.

If I see any, I will definitely make a pull request :)

nemchik (approved):
let's do it

PR #1084, opened 2020-09-03
Validate All Docker Images Exist Every Run

dylanmtaylor:
This tries to use the manifest inspect command to validate the image is there

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -39,3 +39,16 @@ jobs:
           COMMIT_SHA: ${{ github.sha }}
         run: |
           while IFS= read -r line; do echo; sudo -E bash ./main.sh -vt "${line}" || exit 1; echo; done < <(git ls-files '.scripts/*.sh' | sed -E 's/^\.scripts\/(\w+)\.sh$/\1/')
+
+  mainfests:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2.3.2
+      - name: validate
+        working-directory: compose
+        env:
+          CI: true
+          COMMIT_SHA: ${{ github.sha }}
+        run: |
+          dirname -- $(grep -Rl 'deprecated: "true"' .) | xargs rm -rf
+          for image in $(grep -RPoh "image: \K.*" | sort -u | grep -v '${'); do echo "Image: $image" && DOCKER_CLI_EXPERIMENTAL=enabled docker manifest inspect $image; done
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -46,9 +46,6 @@ jobs:
       - uses: actions/checkout@v2.3.2
       - name: Validate Images
         working-directory: compose
-        env:
-          CI: true
-          COMMIT_SHA: ${{ github.sha }}
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/CI.yml
```diff
@@ -39,3 +39,14 @@ jobs:
...
+  mainfests:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2.3.2
+      - name: validate
+        env:
+          CI: true
+          COMMIT_SHA: ${{ github.sha }}
+        run: |
+          for image in $(grep -R "image:" compose | cut -d ":" -f3 | sed 's/^[[:space:]]*//g' | uniq | sort); do echo "Image: $image" && DOCKER_CLI_EXPERIMENTAL=enabled docker manifest inspect $image; done
```
nemchik, line 52:
```suggestion
          for image in $(grep -RPoh "image: \K.*" compose | sort -u); do echo "Image: $image" && DOCKER_CLI_EXPERIMENTAL=enabled docker manifest inspect $image; done
```
is a bit shorter. this is a fair idea but we do have deprecated app files still in the repo mostly for archival purposes. this would require removing those files which i have concerns about. if a user has not updated DS in a very long time and still has that app running when they upgrade DS tries to handle it gracefully and warn the user. removing the apps would result in no warnings and the app would just disappear. technically if a user already has an image pulled they could still run a container with it even after it is pulled from docker hub, however they would loose it if they stopped the container and purged images (functions that DS does have for users to be able to do). I'm open to additional discussion on this though.

dylanmtaylor: I'm working around the deprecation issue now; it shouldn't matter when I'm done.

dylanmtaylor:
Removing the deprecated apps (locally to the CI workflow) is a bit hacky, but it does work. This command could probably be optimized 
`grep -ri 'deprecated: "true"' | cut -f1 -d':' | rev | cut -d'/' -f1 --complement | rev | xargs rm -rf`

.github/workflows/CI.yml
```diff
@@ -39,3 +39,16 @@ jobs:
...
+
+  validate-images:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2.3.2
+      - name: Validate Images
+        working-directory: compose
+        env:
+          CI: true
+          COMMIT_SHA: ${{ github.sha }}
```
nemchik, line 51: I don't think we need the entire `env:` section here. With other CI those are used so that DS scripts can see them.

dylanmtaylor: Yeah, I copied the other ones for consistency. You're right, I'll remove it

.github/workflows/CI.yml
```diff
@@ -39,3 +39,16 @@ jobs:
...
+        run: |
+          dirname -- $(grep -Rl 'deprecated: "true"' .) | xargs rm -rf
```
nemchik, line 53:
```suggestion
          dirname -- $(grep -Rl 'deprecated: "true"') | xargs rm -rf
```
The `.` isn't needed if you're in the directory.

nemchik (approved):
lgtm

PR #1095, opened 2020-09-08
Replaces shared with storage

nemchik:
# Pull request

**Purpose**
`/shared` is a bit ambiguous. Many users would like the benefit of hardlinks and atomic copy/move.

**Approach**
`/storage` should be more obvious what it is to be used for. The intention is to be the parent directory of whatever the user wants in the container that isn't config. Often this may be media or downloads.

There are also some indentation fixes in this branch.

**BREAKING CHANGE**
- Added deprecation notices for `DOWNLOADSDIR`, `DOCKERSHAREDDIR`, and all `MEDIADIR_*` variables/volumes

For those interested in testing:
`ds -u origin/dockerstoragedir`

This will probably merge pretty quick and will come with an announcement on discord regarding the changes.

**Requirements**


angrycuban13 (approved):
From a documentation perspective I haven't found anything wrong.

angrycuban13 (approved):
From a documentation perspective we seem pretty covered.

dylanmtaylor:
This is going to confuse a lot of people, maybe put a note in when they run ds -u saying they're going to need to reconfigure their applications if they are using /data?

nemchik:
> This is going to confuse a lot of people, maybe put a note in when they run ds -u saying they're going to need to reconfigure their applications if they are using /data? It's also going to be tricky for people to reconfigure these apps without losing things like metadata.

I guess i can put `/data` back as well (for now). DS itself isn't really setup to send warnings about deprecated variables. It could be worth looking into though, but the issue is we either send a notice once and it could get overlooked or we send a notice every time certain things happen (like `ds -c` or `ds -u`) but those could become annoying to see so frequently. We're looking at a 3 month period before they are fully removed. I'm open to ideas.

nemchik:
Ok, `/data` is added back (for now) to plex, jellyfin, and emby. They will be removed along with the other deprecated vars/volumes at the end of the year.

PR #1152, opened 2020-10-14
Filebrowser & Docs

### LINKED ISSUE #1145: Adding filebrowser
# Feature request

Adding filebrowser would be a nice asset to this software. It has a bunch of great features such as creating multiple users and adjusting permissions for each user, built in authentication, deleting, moving, copying and renaming files, and more. It already has a Docker image out of the box.

https://github.com/filebrowser/filebrowser

h5ai is good, but its fairly basic and doesn't have as many features as filebrowser. In other words, filebrowser would be a better alternative.

angrycuban13:
# Pull request

**Purpose**
Closes #1145 

**Approach**

**Learning**
Describe the research stage

**Requirements**


compose/.apps/filebrowser/filebrowser.yml
```diff
@@ -0,0 +1,22 @@
+services:
+  filebrowser:
+    container_name: filebrowser
+    environment:
+      - FB_BASEURL=${FILEBROWSER_BASEURL}
+      - FB_DATABASE=${FILEBROWSER_DATABASE}
+      - FB_LOG=${FILEBROWSER_LOG}
+      - FB_PORT=${FILEBROWSER_PORT_80}
+      - PGID=${PGID}
+      - PUID=${PUID}
```
nemchik, line 10: Does filebrowser actually use PUID/PGID?

(Note that I make it a habit to always pass TZ even if it's not used, but I don't do it with PUID/PGID).

compose/.apps/filebrowser/filebrowser.yml
```diff
@@ -0,0 +1,22 @@
+services:
+  filebrowser:
+    container_name: filebrowser
+    environment:
+      - FB_BASEURL=${FILEBROWSER_BASEURL}
+      - FB_DATABASE=${FILEBROWSER_DATABASE}
+      - FB_LOG=${FILEBROWSER_LOG}
+      - FB_PORT=${FILEBROWSER_PORT_80}
```
nemchik, line 8: What is the use case for allowing the port to be set differently inside the container? All I can think of is running the container in host network mode and wanting to change the port. Other than that I would say we should try to keep the app running on port 80 inside the container since it would be an expected port.

p.s. I think qbittorrent is the only app that breaks this rule because it REALLY hates when you set the port on the container without setting the port on the app itself, although that may have changed since it was added.

compose/.apps/filebrowser/filebrowser.ports.yml
```diff
@@ -0,0 +1,4 @@
+services:
+  filebrowser:
+    ports:
+      - ${FILEBROWSER_PORT_80}:${FILEBROWSER_PORT_80}
```
nemchik, line 4: See port comments on the environment section.

compose/.apps/filebrowser/filebrowser.labels.yml
```diff
@@ -0,0 +1,13 @@
...
+      com.dockstarter.appinfo.deprecated: "false"
+      com.dockstarter.appinfo.description: "Filebrowser provides a file managing interface within a specified directory"
+      com.dockstarter.appinfo.nicename: "Filebrowser"
+      com.dockstarter.appvars.filebrowser_base_url: ""
+      com.dockstarter.appvars.filebrowser_database: "/config/database.db"
+      com.dockstarter.appvars.filebrowser_enabled: "false"
+      com.dockstarter.appvars.filebrowser_log: "/config/filebrowser.log"
+      com.dockstarter.appvars.filebrowser_network_mode: ""
+      com.dockstarter.appvars.filebrowser_port_80: "80"
+      com.dockstarter.appvars.filebrowser_share_dir: "~/.config/appdata/filebrowser"
```
nemchik, line 13:
```suggestion
      com.dockstarter.appvars.filebrowser_share_dir: "~/.config/appdata/filebrowser/share"
```
I think this should be a subdirectory to keep it separated from the actual config and not imply that it has to be the same folder as the config.

nemchik (changes requested):

compose/.apps/filebrowser/filebrowser.yml
angrycuban13: I had it on my compose and I didn't have issues as far as I can tell.

I did check on Halian's compose and she passes PGID/PUID as the `user:` so something worth adding and removing this?

compose/.apps/filebrowser/filebrowser.labels.yml
angrycuban13: Fair enough, I just filled that in so it wouldn't fail, but the end user needs to change it to the folder he wants to actually share anyways since that's the path that gets mounted to `/srv`

compose/.apps/filebrowser/filebrowser.yml
nemchik: If the container runs without `user:` i would recommend avoiding it. Running without it means the container runs as root which gives it permissions it might not otherwise have, and for a file browser container I would probably thing having root permissions is a good thing.

nemchik (approved):

PR #1156, opened 2020-10-19
Add Apprise (API)

nemchik:
# Pull request

**Purpose**
Add Apprise API

- [x] Testing: `ds -u origin/apprise`

**Learning**
https://hub.docker.com/r/caronc/apprise
https://github.com/caronc/apprise-api

**Requirements**


angrycuban13 (approved):

nemchik:
I tried running this and had some odd permissions issues. Hoping someone else can clarify if they have the same or if there's something I need to adjust in the template before merging.

Xyv:
Hi,

I think I managed to "solve" the problems to include this app. Although, the solutions for now are not satisfactory since they require user intervention: there is probably the need to exchange with the package maintainer for a proper fix.

I've stumbled on two problems trying to make this app work:
1. Permissions issue on the mounted config folder: turns out the app uses the `www-data` user (default django/nginx user from what I gathered). I don't think this is configurable, for now, through the `PUID`/`PGID` the linuxserver images uses. 
    **Solution**: A `chown www-data:www-data` on the mounted config folder "solves" this for now.

2. Trying to connect from outside the local network behind a reverse proxy (using `swag`) I got a `400 Bad Request Contradictory scheme headers`, which is a `gunicorn` error that led me to benoitc/gunicorn#1857. 
  The problem arises from the fact that swag forward the request as `https` (From `proxy_set_header X-Forwarded-Proto https;` in proxy.conf)  while connecting to the app through `http` (I made a nginx conf file from _template.subdomain.conf.sample):
 ```nginx
        set $upstream_app apprise;
        set $upstream_port 8000;
        set $upstream_proto http;
        proxy_pass $upstream_proto://$upstream_app:$upstream_port;
``` 
  `gunicorn` then sees a mismatch between the accessed protocol (http) and the forwarded protocol (https) and throws the error. 
    **Solution**: Add the line `secure_scheme_headers = {'X-FORWARDED-PROTO': 'http'}` in the gunicorn configuration file (Which need to be copied from the container in `/opt/apprise/webapp/gunicorn.conf.py`, modified then mounted in `docker-compose.yml` ...) 
  There is probably a better way of solving this, but given my limited understanding of the problem, it is working. (Such as passing this info to gunicorn through env variable maybe).

nemchik:
Switched to LSIO's image. Tested and ready.

Issue #740, opened 2019-06-26
Better Logging


nemchik:
Remaining todo:

> Be more verbose with `info`, for example have the start and end of each function announce itself or something

---

> https://stackoverflow.com/a/11905720/1384186 
> Allow redirecting stdin to log functions

Decided not to do this

> Add additional log functions

Done (`trace`, `debug`, and `notice`) in #765

> Add the ability to specify a log level

Done (`-v` and `-x`) in #765

> `prompt` probably needs to be it's own log functions that cannot be turned off

Done (`notice`) in #765

nemchik:
Good ideas here https://github.com/idelsink/b-log

nemchik:
Interesting but different https://github.com/nk412/tinylogger

nemchik:
https://github.com/GhostWriters/DSAC/commit/7f46c487cb2ddf2da07e5ed8c2ddff08f625c0c8#commitcomment-33417543

nemchik:
Keep

nemchik:
We can still finish improving this.

nemchik:
More `info` wouldn't hurt, but I'm happy to close this with the new `fatal` additions.

PR #1174, opened 2020-11-06
Output failing commands with fatals

nemchik:
# Pull request

**Purpose**
Closes https://github.com/GhostWriters/DockSTARTer/issues/740
Display the command that was run that caused DS to fatal and exit. This way users can try running that command on their own, or report it to us for support.

**Approach**
I couldn't find a better way to do this other than literally copying the commands and putting them in the output. I swear there has to be a better way. but this might be all we can do for now. It's not very pretty.

**Requirements**


nemchik:
Tried https://github.com/GhostWriters/DockSTARTer/commit/01a9fd0334eb5356416c7397dd2eddc37772cb74 unsuccessfully.

PR #1192, opened 2020-12-04
Bitwarden

angrycuban13:
# Pull request

**Purpose**
Removes unnecessary variables from when you first enable Bitwarden. When first entering the admin page it says the following:

```
NOTE: The settings here override the environment variables. 
Once saved, it's recommended to stop setting them to avoid confusion. 
This does not apply to the read-only section, which can only be set through the environment. 
```

Some of the settings in the admin page are the SMTP settings.

The way the container is set up it does not let you have empty variables. Removing these variables from the compose file will help users get started just a bit faster.


angrycuban13:
Ummm let me close this and re-open it. It looks like my working directory is a little dirty.

PR #1230, opened 2021-01-09
Rewrite env handling, attempt to prevent accidental clean slate

nemchik:
# Pull request

**Purpose**
After some reports on Discord of `.env` files being reset to default I located the code in DS that would replace a user's `.env` with the `.env.example` which happens in `.scripts/env_update.sh`. This was done in order to ensure the user's `.env` file is kept up to date with the DS example file. The original process was to copy the user's `.env` to a temporary file and sort it alphabetically, then remove the original `.env` and copy the example to the original location, then write all the values from the temporary `.env` to the new copy. The issue here was replacing the original `.env` with the example and then writing the temporary file values back could sometimes fail leaving users with a stock `.env`.

**Approach**
Rewrite `.scripts/env_update.sh` to:

- Copy the current `.env` values to a temporary file and sort them alphabetically (unchanged from previous behavior)
- Copy the example `.env` file to a different temporary file
- Write the values from the temporary sorted file into the temporary example file (referred to below as temporary completed file)
- Copy the temporary completed file directly on top of the original `.env` using `cp -f`
- Remove both temporary files and set permissions on the original `.env` file (unchanged from previous behavior)

This process required adding the ability to specify the path of the `.env` file in `.scripts/env_set.sh` for which file the values should be written to. I also added it to `.scripts/env_get.sh` for consistency.

- [x] Testing `ds -u origin/env`

**Requirements**


nemchik:
Have not had any major issues reported. Ready to merge.

PR #1240, opened 2021-01-16
Add MusicBrainz Picard to DS

D4rkSl4ve:
Add MusicBrainz Picard to the DockerSTARTer list of apps.


angrycuban13 (changes requested):
@D4rkSl4ve can you please also add a page for the documentation? It can be located in `docs/apps` and you'd just create a new file called `picard.md`. Look at how the rest of the `md` files are structured so you get a general idea. Other than that, it looks great!

angrycuban13 (approved):

PR #1289, opened 2021-02-21
Add missing volume for cloudflareddns

nemchik:
Using
```shell
docker ps -a --format '{{ .ID }}' | xargs -I {} docker inspect -f '{{ .Name }}{{ printf "\n" }}{{ range .Mounts }}{{ printf "\n\t" }}{{ .Type }} {{ if eq .Type "bind" }}{{ .Source }}{{ end }}{{ .Name }} => {{ .Destination }}{{ end }}{{ printf "\n" }}' {}
```
to find volumes not being created as bind. Each one may need to be handled differently.


MagicalCodeMonkey (approved):
I just noticed this a little bit ago and came here to make a PR 😄 
Looks good to me!

PR #1301, opened 2021-03-20
Overrides for Petio

TRaSH-:
Overrides for Petio thnx to mouzzampk2014

# Pull request

**Purpose**
Added Override for Petio

**Approach**
Added Override for Petio

**Learning**
Describe the research stage

**Requirements**

### DIFF
--- a/docs/overrides/petio.md
+++ b/docs/overrides/petio.md
@@ -15,27 +15,37 @@
 services:
 
   petio:
-    image: ghcr.io/hotio/petio
-    hostname: ${DOCKERHOSTNAME}
     container_name: petio
     environment:
-      - PGID=${PGID}
-      - PUID=${PUID}
       - TZ=${TZ}
+    hostname: ${DOCKERHOSTNAME}
+    image: ghcr.io/hotio/petio
+    logging:
+      driver: json-file
+      options:
+        max-file: ${DOCKERLOGGING_MAXFILE}
+        max-size: ${DOCKERLOGGING_MAXSIZE}
     ports:
-      - "7777:7777"
+      - 7777:7777
     restart: unless-stopped
     volumes:
       - /etc/localtime:/etc/localtime:ro
       - ${DOCKERCONFDIR}/petio:/config
   mongo:
-    image: mongo
-    hostname: ${DOCKERHOSTNAME}
     container_name: mongo
-    user: "1000:1000"
     environment:
       - TZ=${TZ}
+    hostname: ${DOCKERHOSTNAME}
+    image: mongo
+    logging:
+      driver: json-file
+      options:
+        max-file: ${DOCKERLOGGING_MAXFILE}
+        max-size: ${DOCKERLOGGING_MAXSIZE}
+    ports:
+      - 27017:27017
     restart: unless-stopped
+    user: ${PUID}:${PGID}
     volumes:
       - /etc/localtime:/etc/localtime:ro
       - ${DOCKERCONFDIR}/mongo:/data/configdb
--- a/docs/overrides/petio.md
+++ b/docs/overrides/petio.md
@@ -1,0 +1,43 @@
+# Petio
+
+[![Docker Pulls](https://img.shields.io/docker/pulls/hotio/petio?style=flat-square&color=607D8B&label=docker%20pulls&logo=docker)](https://hub.docker.com/r/hotio/petio)
+[![GitHub Stars](https://img.shields.io/github/stars/hotio/petio?style=flat-square&color=607D8B&label=github%20stars&logo=github)](https://github.com/hotio/petio)
+
+## Description
+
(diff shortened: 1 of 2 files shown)


docs/overrides/petio.md
```diff
@@ -0,0 +1,43 @@
+# Petio
+
+[![Docker Pulls](https://img.shields.io/docker/pulls/hotio/petio?style=flat-square&color=607D8B&label=docker%20pulls&logo=docker)](https://hub.docker.com/r/hotio/petio)
+[![GitHub Stars](https://img.shields.io/github/stars/hotio/petio?style=flat-square&color=607D8B&label=github%20stars&logo=github)](https://github.com/hotio/petio)
+
+## Description
+
+[petio](https://petio.tv/) is a third party companion app available to Plex server owners to allow their users to request, review and discover content. The app is built to appear instantly familiar and intuitive to even the most tech-agnostic users. Petio will help you manage requests from your users, connect to other third party apps such as Sonarr and Radarr, notify users when content is available and track request progress. Petio also allows users to discover media both on and off your server, quickly and easily find related content and review to leave their opinion for other users.
```
angrycuban13, line 8: Capitalize Petio

docs/overrides/petio.md
```diff
@@ -0,0 +1,43 @@
...
+    image: mongo
+    hostname: ${DOCKERHOSTNAME}
+    container_name: mongo
+    user: "1000:1000"
+    environment:
+      - TZ=${TZ}
+    restart: unless-stopped
+    volumes:
+      - /etc/localtime:/etc/localtime:ro
+      - ${DOCKERCONFDIR}/mongo:/data/configdb
```
angrycuban13, line 41: This line isn't really needed. I think Hotio added it for a reason but he directs you to read the docs. It won't really apply to the majority of users in my experience

angrycuban13 (changes requested):

docs/overrides/petio.md
TRaSH-: the confdir ?
are you sure

angrycuban13: The one where the database lives is actually the line below. 

```The folder /data/configdb isn't always required, take a look at the Mongo documentation to see if this applies to you.```

https://hotio.dev/containers/petio/

I think it would create less confusion for users if we don't include it but up to you. Especially since Hotio is directing them to create the folders *beforehand*.

Comment (mouzzampk2014) [reply docs/overrides/petio.md:41]: I think this whole override could be more simple. I just posted it in a hurry so people will take advantage of petio. 

As all our apps include config directory so might be ideal to keep it when we add in the main repo but for override I think its better to remove the configdb. 

We need to add :rw to give permissions to folder otherwise users need to first create a folder and then set permission and then run image but adding :rw to config will avoid the steps.

Comment (TRaSH-) [reply docs/overrides/petio.md:41]: i think personally we should just add it, this way when you got users that actually RTFM they come over and complain we're missing something.

Comment (nemchik) [reply docs/overrides/petio.md:41]: The mongo folder should be inside the petio config folder

Comment (nemchik) [reply docs/overrides/petio.md:41]: Ignore me. I didn't read the whole thing.

Comment (TRaSH-) [reply docs/overrides/petio.md:41]: so something like this ?

```yml
  mongo:
    image: mongo
    hostname: ${DOCKERHOSTNAME}
    container_name: mongo
    user: "1000:1000"
    environment:
      - TZ=${TZ}
    restart: unless-stopped
    volumes:
      - /etc/localtime:/etc/localtime:ro
      - ${DOCKERCONFDIR}/petio/mongo:/data/configdb
      - ${DOCKERCONFDIR}/petio/mongo/db:/data/db
```

TRaSH-: > Ignore me. I didn't read the whole thing.

to late it already happened

HalianElf: Shouldn't it have `user: "${PUID}:${PGID}"` instead of 1000:1000?

angrycuban13: Yep

angrycuban13: > Ignore me. I didn't read the whole thing.

Ignored

angrycuban13:
I'll sit on a computer and take a look in a min.

docs/overrides/petio.md
angrycuban13: So per Mongo's [documentation](https://docs.mongodb.com/manual/reference/configuration-options/#configuration-file) 

```You can configure mongod and mongos instances at startup using a configuration file. The configuration file contains settings that are equivalent to the mongod and mongos command-line options.```

I think this is more advanced usage of Mongo than our average users. It might just create confusion but I don't care if it gets merged with this line.

File: docs/overrides/petio.md [hunk 3]
```diff
@@ -0,0 +1,43 @@
...
+      - "7777:7777"
+    restart: unless-stopped
+    volumes:
+      - /etc/localtime:/etc/localtime:ro
+      - ${DOCKERCONFDIR}/petio:/config
+  mongo:
+    image: mongo
+    hostname: ${DOCKERHOSTNAME}
+    container_name: mongo
+    user: "1000:1000"
```
angrycuban13, line 35: Change to

`user: ${PUID}:${PGID}`

angrycuban13 (commented):
Petio also supports

```yml
user: ${PUID}:${PGID}
```

@nemchik should we move away from environment variables if the `user` option is supported?

angrycuban13 (approved):

PR #1321, opened 2021-06-05
Use pm to install docker

nemchik:
# Pull request

**Purpose**

**Approach**

**Learning**
Describe the research stage

**Requirements**


nemchik:
CentOS installs `podman` when running `yum install docker`. podman seems to not play nice. We either need to adjust DS to work with podman or find a way to get vanilla docker to install (preferable).

nemchik:
Reconsidered and reworked this. The `pm` script structure can be used to install docker, however since `apt` applies to Ubuntu and Debian (and more) and `dnf` applies to Fedora and CentOS8+ (CentOS7 uses `yum`) it's not reliable to check for the existence of a package manager and install packages without knowing the repository will have them. The https://get.docker.com/ script seems to be fairly reliable at performing this correctly. The primary concern with using the https://get.docker.com/ script is at times a new version of Ubuntu will release and the official docker repo will not have a release readily available for weeks or even sometimes months. The best action may be to continue advising users to install docker manually in those cases or wait for an official `docker-ce` release.

This thought is partly motivated by docker's announcement that docker compose v2 will in the near future be baked into the docker engine. This would likely mean that the official docker repos would be the best way to install docker with compose. This may not always be the case (long term) but at least for the time being.

So why do anything at all?
This PR reorganizes how DS treats the installation of docker. The primary difference is that `pacman`/Arch can be fully supported, and additional package manager types (ex: `apk`/alpine) can be added in the future.

Additional thoughts:
CentOS8 will be EOL at the end of the year. CentOS7 will somehow not be EOL until 2024. CentOS8 uses `dnf` and thus conflicted with Fedora's ability to install docker from the package manager. Once CentOS8 is fully EOL I may revisit this in some ways. `apt` systems will likely always be a challenge as described above, and thus relying on https://get.docker.com/ may be a requirement for DS.

PR #1502, opened 2022-07-26
Add RustDesk support (WIP - needs ports still)

dylanmtaylor:


shahvirb:
Hey you might be interested in my [Python project ](https://github.com/shahvirb/dockstarter-dev-tools) to help generate these dockstarter app yml files. I don't have a readme but the premise is that you can pass in a working docker compose yml and it will generate the needed dockstarter app yml files by splitting the relevant sections from your source docker compose yml.

shahvirb:
Check out this section I added lately. Would be interested in your thoughts/feedback as you are adding this app as to what I've missed.

https://github.com/GhostWriters/DockSTARTer/blob/master/.github/CONTRIBUTING.md#adding-an-app

dylanmtaylor:
> Check out this section I added lately. Would be interested in your thoughts/feedback as you are adding this app as to what I've missed.
> 
> https://github.com/GhostWriters/DockSTARTer/blob/master/.github/CONTRIBUTING.md#adding-an-app

This is actually super helpful, thank you! I'm going to go ahead and test this. The shell check failure isn't in something I'm adding, so I'm ignoring that for now.

dylanmtaylor:
> Check out this section I added lately. Would be interested in your thoughts/feedback as you are adding this app as to what I've missed.
> 
> https://github.com/GhostWriters/DockSTARTer/blob/master/.github/CONTRIBUTING.md#adding-an-app

Maybe you can help me with this one -- in their Docker example it looks like there are two containers in the compose file. Any clean way to handle this in the DockStarter yml config?

https://rustdesk.com/docs/en/self-host/install/

shahvirb:
> in their Docker example it looks like there are two containers in the compose file. Any clean way to handle this in the DockStarter yml config?

Unfortunately we don't have multi-container app support in dockstarter. So you need to add this as two separate apps and then in the app specific docs we should tell the user to install both apps in dockstarter.

dylanmtaylor:
Well, this part didn't work.

What would you like set for RUSTDESK_HBBR_HOST?     │ 
                                                                                          │                                                     │ 
                                                                                          │           Keep Current  ${DOCKERHOSTNAME}           │ 
                                                                                          │           Use Default                               │ 
                                                                                          │           Enter New                                 │ 
                                                                                          │                                                     │

dylanmtaylor:
It's actually working! :)

image

I put the real key and IP for my server in, and it shows the green ready indicator on the bottom. With a bad IP or key, it shows an orange indicator and shows it as trying to connect to the network. localhost works if both are on the same system, otherwise you can point at the hbbr server on a second machine.

dylanmtaylor:
@nemchik I believe that this is in a state where it's usable and complete. Would you please review this pull request?

dylanmtaylor:
So I just found https://github.com/rustdesk/rustdesk-server#s6-overlay-based-images -- we could very likely combine these back into one app after all.

dylanmtaylor:
Surprisingly, I removed the split apps and tested the commit I just made and it still works. This is honestly much cleaner.

compose/.apps/rustdesk/rustdesk.labels.yml
```diff
@@ -0,0 +1,15 @@
+services:
+  rustdesk:
+    labels:
+      com.dockstarter.appinfo.deprecated: "false"
+      com.dockstarter.appinfo.description: The ID/rendezvous server for RustDesk, an open-source self-hosted TeamViewer alternative
```
shahvirb, line 5:
```suggestion
com.dockstarter.appinfo.description: RustDesk server software including hbbs (rendezvous server) and hbbr (relay server) program, an open-source self-hosted TeamViewer alternative
```

compose/.apps/rustdesk/rustdesk.yml
```diff
@@ -0,0 +1,17 @@
...
+      - PUID=${PUID}
+      - TZ=${TZ}
+      - RELAY=${RUSTDESK_RELAY_SERVER}
+    logging:
+      driver: json-file
+      options:
+        max-file: ${DOCKERLOGGING_MAXFILE}
+        max-size: ${DOCKERLOGGING_MAXSIZE}
+    restart: ${RUSTDESK_RESTART}
+    volumes:
```
shahvirb, line 15: ```- ${DOCKERSTORAGEDIR}:/storage``` is always included
See https://github.com/GhostWriters/DockSTARTer/blob/master/.github/CONTRIBUTING.md

shahvirb:
Looks mostly good to me. Two small things remaining. See my suggestions.

dylanmtaylor:
> Looks mostly good to me. Two small things remaining. See my suggestions.

Do you have a review in a pending state or something? I can't see them. Could you elaborate a bit on what I should improve?

shahvirb (changes requested):

compose/.apps/rustdesk/rustdesk.yml
dylanmtaylor: Is there any reason to grant the server access to my storage directory if it will never use it? Just seems to be poor security posture (least needed access, etc.). If it is required by the project I can make the change, but in this specific case, it feels unwise.

compose/.apps/rustdesk/rustdesk.aarch64.yml
```diff
@@ -0,0 +1,3 @@
+services:
+  rustdesk:
+    image: rustdesk/rustdesk-server-s6:latest
```
nemchik, line 3:
```suggestion
    image: rustdesk/rustdesk-server-s6
```
`latest` is implied, and thus not needed.

compose/.apps/rustdesk/rustdesk.armv7l.yml
```diff
@@ -0,0 +1,3 @@
+services:
+  rustdesk:
+    image: rustdesk/rustdesk-server-s6:latest
```
nemchik, line 3:
```suggestion
    image: rustdesk/rustdesk-server-s6
```
`latest` is implied, and thus not needed.

compose/.apps/rustdesk/rustdesk.x86_64.yml
```diff
@@ -0,0 +1,3 @@
+services:
+  rustdesk:
+    image: rustdesk/rustdesk-server-s6:latest
```
nemchik, line 3:
```suggestion
    image: rustdesk/rustdesk-server-s6
```
`latest` is implied, and thus not needed.

compose/.apps/rustdesk/rustdesk.yml
```diff
@@ -0,0 +1,18 @@
+services:
+  rustdesk:
+    container_name: rustdesk
+    environment:
+      - PGID=${PGID}
+      - PUID=${PUID}
+      - TZ=${TZ}
+      - RELAY=${RUSTDESK_RELAY_SERVER}
```
nemchik, line 8:
```suggestion
      - RELAY=${RUSTDESK_RELAY_SERVER}
      - TZ=${TZ}
```
Sort these alphabetically please.

nemchik (changes requested):
This looks great! A few small things to adjust and it'll be ready to merge. Thank you!

dylanmtaylor:
> This looks great! A few small things to adjust and it'll be ready to merge. Thank you!

I've made the requested corrections :)

nemchik:
I updated your branch with changes from master to fix the shellcheck linting.

dylanmtaylor:
> I updated your branch with changes from master to fix the shellcheck linting.

Thanks, that's fine with me. This is ready to be merged as soon as you're happy with it :)

nemchik (approved):
LGTM

Issue #1448, opened 2022-03-28
[Application Request] Deprecate Airsonic in favour of Airsonic-advanced


SevereOverfl0w:
# Name of the application

airsonic-advanced

## GitHub

https://github.com/airsonic-advanced/airsonic-advanced

## Docker Hub/GitHub Registry

https://github.com/linuxserver/docker-airsonic-advanced

### Additional context

From [linuxserver/airsonic](https://hub.docker.com/r/linuxserver/airsonic):

> This image is deprecated. We will not offer support for this image and it will not be updated. We recommend our airsonic-advanced image instead: https://github.com/linuxserver/docker-airsonic-advanced

nemchik:
This does need to happen. I haven't had time. Could be as simple as switching the image in the compose, but we need to make sure it's compatible with the existing vars/ports/volumes (etc) or if we need new compose.

PR #1453, opened 2022-04-02
Replacing Airsonic with Airsonic Advanced

angrycuban13:
# Pull request

**Purpose**
Closes #1448

PR #1504, opened 2022-07-28
Switch to Airsonic-Advanced

nemchik:
# Pull request

**Purpose**
This closes https://github.com/GhostWriters/DockSTARTer/issues/1448

**Requirements**

### DIFF
--- a/compose/.apps/airsonic/airsonic.aarch64.yml
+++ b/compose/.apps/airsonic/airsonic.aarch64.yml
@@ -1,3 +1,3 @@
 services:
   airsonic:
-    image: lscr.io/linuxserver/airsonic
+    image: lscr.io/linuxserver/airsonic-advanced
--- a/compose/.apps/airsonic/airsonic.armv7l.yml
+++ b/compose/.apps/airsonic/airsonic.armv7l.yml
@@ -1,3 +1,3 @@
 services:
   airsonic:
-    image: lscr.io/linuxserver/airsonic
+    image: lscr.io/linuxserver/airsonic-advanced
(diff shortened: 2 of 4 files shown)


shahvirb:
@nemchik I have mixed feelings about this approach of just updating the image to "-advanced"

https://github.com/linuxserver/docker-airsonic-advanced#application-setup says "We don't formally support upgrading from Airsonic to Airsonic Advanced, it may or may not work for you and we'd recommend making backups before attempting this."

I think the safer approach would be to create a new "airsonic-advanced" app and we mark the "airsonic" app as deprecated.

CLHatch:
> @nemchik I have mixed feelings about this approach of just updating the image to "-advanced"
> 
> https://github.com/linuxserver/docker-airsonic-advanced#application-setup says "We don't formally support upgrading from Airsonic to Airsonic Advanced, it may or may not work for you and we'd recommend making backups before attempting this."
> 
> I think the safer approach would be to create a new "airsonic-advanced" app and we mark the "airsonic" app as deprecated.

Doing it that way, if someone wanted to try to transfer their old setup over to the "advanced" one, they could just rename their `~/.config/appdata/airsonic` folder to `~/.config/appdata/airsonnic-advanced` anyways, knowing that it MIGHT not work (I'm just assuming those are the folder names, not looked).

shahvirb:
Yes indeed. I think this is the safer approach but obviously less
convenient for users as a simple ds -u won't just update their app to the
advanced image.

Given that airsonics position is "proceed at your own risk" I nominate that
we respect this and don't force upgrade our users.

nemchik:
I can appreciate all of that. In that case, I think reopening https://github.com/GhostWriters/DockSTARTer/pull/1453 makes more sense. The reason I went this direction was to avoid deprecation of the current image just to make things easier on users, but if the safer approach is to deprecate and make users migrate I'm fine with that. We should cover it in the docs.

PR #1535, opened 2022-09-16
Added support for TRANSMISSION_WEB_HOME environment variable

marcellus00:
# Pull request

**Purpose**
Linuxcontainer transmission image contains a few alternative web ui's, however it was impossible to set an environment variable TRANSMISSION_WEB_HOME to use them

**Approach**
Simple fix by adding TRANSMISSION_WEB_HOME to transmission[vpn][.labels].yml files

**Learning**

**Requirements**


compose/.apps/transmissionvpn/transmissionvpn.yml
```diff
@@ -21,6 +21,7 @@ services:
       - TRANSMISSION_HOME=/config
       - TRANSMISSION_INCOMPLETE_DIR=${TRANSMISSIONVPN_INCOMPLETE_DIR}
       - TRANSMISSION_WATCH_DIR=${TRANSMISSIONVPN_WATCH_DIR}
+      - TRANSMISSION_WEB_HOME=${TRANSMISSIONVPN_WEB_HOME}
```
shahvirb, line 24: Does this image support this variable? Did you do a functional test?

compose/.apps/transmission/transmission.labels.yml
```diff
@@ -10,3 +10,4 @@ services:
       com.dockstarter.appvars.transmission_port_6881: "6881"
       com.dockstarter.appvars.transmission_port_9091: "9091"
       com.dockstarter.appvars.transmission_restart: unless-stopped
+      com.dockstarter.appvars.transmission_web_home: ""
```
shahvirb, line 13: What happens when this variable is blank? Does the app default to the standard UI?

compose/.apps/transmissionvpn/transmissionvpn.yml
nemchik: These variables are not specific to the image, they are supported by Transmission. https://github.com/transmission/transmission/blob/main/docs/Environment-Variables.md

This would be a correct implementation.

nemchik (changes requested):
I'm OK with these changes, but I would like to also see information added to the documentation files (in the docs folder) explaining what you might set these variables to. I believe at some point in the past, binhex's transmissionvpn image included alternate transmission web UIs, so it might have been possible to use this variable to access them. I don't believe the lsio transmission image has ever included alternate UIs, so it would be good to know how to use the variable. My guess is you need to download the alternate web UI separately, put it in your config folder (or somewhere the container can see) and use a path like `/config/ui/alt` or something.

PR #1626, opened 2023-04-19
Update airsonicadvanced.labels.yml

SpiderManAPV:
Fixes the airsonic-advanced config to correctly list and pull the proper image rather than mixing airsonic with airsonic-advanced.

# Pull request

**Purpose**
As it stands right now, the nicename for Airsonic and Airsonic-Advanced is the same. That leads to DS incorrectly pulling the original airsonic rather than advanced.

**Approach**
Changes the nicename for AirSonic-Advanced to be different from Airsonic.

N/A

**Learning**
N/A

**Requirements**


nemchik (approved):
Good catch. I removed the dash because DS is built for the `nicename` to match exactly the variable prefix, which cannot have dashes.

Issue #1635, opened 2023-04-30
Update "murmur" image


diogovbsantos:
Hello,

The current "murmur" implementation uses an image with a very outdated server version (not updated in 2 years).

I propose changing the image to mumblevoip/mumble-server, which is the official one - and has the latest update.

PR #1662, opened 2023-08-09
Switch murmur container to mumble-voip/mumble-docker

dylanmtaylor:
# Pull request

**Purpose**
This is a work in progress that I'm submitting for comments. It should address #1635.

**Approach**

This replaces the image while trying to keep the old variables the same so things don't break on upgrade for users.


nemchik:
Sorry for taking so long to get to this. PR looks good to me, but based on the volume changes I have a question. Will the existing user data be used by the new container? or will users have to shuffle around some of the files in their config folder to have the new container pick them up?

dylanmtaylor:
> Sorry for taking so long to get to this. PR looks good to me, but based on the volume changes I have a question. Will the existing user data be used by the new container? or will users have to shuffle around some of the files in their config folder to have the new container pick them up?

I'm doubtful that data will transfer over cleanly.

nemchik:
I do like the idea of switching to this image, but I also wonder if we should just make a new app folder with the name `mumbleserver` and deprecate this one as the image hasn't been updated in years. There are a few additional environment vars like the UID/GID that we could also include with the new image. I'm also working on a pretty insanely massive PR with #1729 that has breaking changes (which I try to mitigate as much as possible) for end users, including a handful of deprecations (and new apps). So maybe it could be tucked into that, or wait until that merges since the naming cadence for variables (and thus labels) is changing.

dylanmtaylor:
> I do like the idea of switching to this image, but I also wonder if we should just make a new app folder with the name `mumbleserver` and deprecate this one as the image hasn't been updated in years. There are a few additional environment vars like the UID/GID that we could also include with the new image. I'm also working on a pretty insanely massive PR with #1729 that has breaking changes (which I try to mitigate as much as possible) for end users, including a handful of deprecations (and new apps). So maybe it could be tucked into that, or wait until that merges since the naming cadence for variables (and thus labels) is changing.

If you want to work this image change into your breaking PR I'm fine with that. I'm honestly not sure how we can do this in a way which is least annoying to users of mumble. I was trying to seamlessly swap over the container so it gets recreated without them having to redo their configuration but the data structure is different.

PR #1726, opened 2024-02-10
Added some info for the docs on how to access WebUI

psan-dev:
# Pull request

**Purpose**
qBittorrent randomly generates a password for WebUI which is only printed to stdout upon startup. This may pose an impediment to some users.

**Approach**
Added documentation to guide users on how to successfully access WebUI

**Learning**
Own research

**Requirements**

### DIFF
--- a/docs/apps/qbittorrent.md
+++ b/docs/apps/qbittorrent.md
@@ -20,9 +20,9 @@ you need assistance setting up this application please visit our
 
 qBittorrent randomly generates a password for the WebUI login at startup until
 one has been set by the user. This means when you first setup the container you
-wont be able to access WebUI until you find out the randomly generated password
+wont be able to access WebUI until you find out the randomly generated password.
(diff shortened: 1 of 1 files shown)


nemchik:
Could you instead run `docker logs qbittorrent` after starting the container for the first time? That should show the password without the need to run two shells and kill the process while watching the output.

psan-dev:
> Could you instead run `docker logs qbittorrent` after starting the container for the first time? That should show the password without the need to run two shells and kill the process while watching the output.

Yeah, this is fair. That's probably a better suggestion.

nemchik:
> > Could you instead run `docker logs qbittorrent` after starting the container for the first time? That should show the password without the need to run two shells and kill the process while watching the output.
> 
> Yeah, this is fair. That's probably a better suggestion.

If you'd like to add information about viewing the temp password from the logs to the documentation that would be accepted.

PR #1731, opened 2024-02-20
Make DockStarter work with CoreOS (immutable filesystem, no traditional package manager)

### LINKED ISSUE #1730: [Feature Request] Ability to install in CoreOS
# Feature request

It doesn't install on CoreOS (specifically the [ucore-hci](https://github.com/ublue-os/ucore) spin, as the filesystem is immutable)
```
2024-02-20 19:26:28 [INFO  ]   Creating /usr/bin/ds symbolic link for DockSTARTer.
ln: failed to create symbolic link '/usr/bin/ds': Read-only file system
2024-02-20 19:26:28 [FATAL ]   Failed to create symlink.
Failing command: sudo ln -s -T "/var/home/dylan/.docker/main.sh" /usr/bin/ds
DockSTARTer did not finish running successfully.
```

Make the symlink at $HOME/bin/ instead if the root filesystem is read only.

N/A

**Additional context**
N/A

dylanmtaylor:
# Pull request

**Purpose**
Closes #1730. 

Tested on [ucore-hci from the Universal Blue project](https://github.com/ublue-os/ucore). Manual workaround needed for docker compose plugin:
```
DOCKER_CONFIG=${DOCKER_CONFIG:-$HOME/.docker}
mkdir -p $DOCKER_CONFIG/cli-plugins
curl -SL https://github.com/docker/compose/releases/download/v2.24.5/docker-compose-linux-x86_64 -o $DOCKER_CONFIG/cli-plugins/docker-compose
chmod +x $DOCKER_CONFIG/cli-plugins/docker-compose
```
I also set SELinux to permissive. After that, it worked perfectly and all my containers spun up. :-)
image

**Approach**
* Uses $HOME/bin instead of /usr for the filesystem
* Doesn't fail hard if it detects no supported package manager, instead checking if the commands we need are there.

**Learning**
Describe the research stage

**Requirements**

### DIFF
--- a/.scripts/package_manager_run.sh
+++ b/.scripts/package_manager_run.sh
@@ -26,6 +26,11 @@ package_manager_run() {
             fi
         done
 
+        if [[ "$(docker compose 2>&1)" == *"docker: 'compose' is not a docker command."* ]]; then
+            fatal "The 'docker compose' command is not functional. Follow the directions at https://docs.docker.com/compose/install/linux/ to install compose."
+        fi
+
+
     fi
 }
 
--- a/.scripts/package_manager_run.sh
+++ b/.scripts/package_manager_run.sh
@@ -12,7 +12,7 @@ package_manager_run() {
         run_script "pm_pacman_${ACTION}"
     elif [[ -n "$(command -v yum)" ]]; then
         run_script "pm_yum_${ACTION}"
-    elif [[ "${ACTION}" == "install" ]]; then
+    elif [[ ${ACTION} == "install" ]]; then
         # We might not need a supported package manager at all if the dependencies are there already. Let's validate that.
 
         # Define an array of commands
--- a/.scripts/package_manager_run.sh
+++ b/.scripts/package_manager_run.sh
@@ -14,7 +14,7 @@ package_manager_run() {
         run_script "pm_yum_${ACTION}"
     else
         # We might not need a supported package manager at all if the dependencies are there already. Let's validate that.
-        echo "Supported package manager not detected. Checking for dependencies ..."
+
         # Define an array of commands
         commands=("curl" "docker" "docker-compose" "git" "grep" "sed" "whiptail")
 
@@ -26,7 +26,6 @@ package_manager_run() {
             fi
         done
 
-        echo "All commands are available."
     fi
 }
 
--- a/.scripts/package_manager_run.sh
+++ b/.scripts/package_manager_run.sh
@@ -16,7 +16,7 @@ package_manager_run() {
         # We might not need a supported package manager at all if the dependencies are there already. Let's validate that.
 
         # Define an array of commands
-        commands=("curl" "docker" "git" "grep" "sed" "whiptail")
+        commands=("curl" "git" "grep" "sed" "whiptail")
 
         # Iterate over each command in the array
         for cmd in "${commands[@]}"; do
@@ -25,7 +25,13 @@ package_manager_run() {
                 fatal "Error: '$cmd' is not available. Exiting..."
             fi
         done
+    elif [[ ${ACTION} == "install_docker" ]]; then
+        # Check for the presence of the docker command
+        if ! command -v "docker" &> /dev/null; then
+            fatal "Error: 'docker' is not available. Exiting..."
+        fi
 
+        # If docker warns that compose is not a docker command when we call it, we alert the user they need to take action.
         if [[ "$(docker compose 2>&1)" == *"docker: 'compose' is not a docker command."* ]]; then
             fatal "The 'docker compose' command is not functional. Follow the directions at https://docs.docker.com/compose/install/linux/ to install compose."
         fi
--- a/.scripts/package_manager_run.sh
+++ b/.scripts/package_manager_run.sh
@@ -32,7 +32,7 @@ package_manager_run() {
         fi
 
         # If docker warns that compose is not a docker command when we call it, we alert the user they need to take action.
-        if [[ "$(docker compose 2>&1)" == *"docker: 'compose' is not a docker command."* ]]; then
+        if ! docker compose version > /dev/null 2>&1; then
             fatal "The 'docker compose' command is not functional. Follow the directions at https://docs.docker.com/compose/install/linux/ to install compose."
         fi
     fi
--- a/.scripts/package_manager_run.sh
+++ b/.scripts/package_manager_run.sh
@@ -13,7 +13,20 @@ package_manager_run() {
     elif [[ -n "$(command -v yum)" ]]; then
         run_script "pm_yum_${ACTION}"
     else
-        fatal "Supported package manager not detected!"
+        # We might not need a supported package manager at all if the dependencies are there already. Let's validate that.
+        echo "Supported package manager not detected. Checking for dependencies ..."
+        # Define an array of commands
+        commands=("curl" "docker" "docker-compose" "git" "grep" "sed" "whiptail")
+        
+        # Iterate over each command in the array
+        for cmd in "${commands[@]}"; do
+            # Check if the command is available in the system
+            if ! command -v "$cmd" &> /dev/null; then
+                fatal "Error: '$cmd' is not available. Exiting..."
+            fi
+        done
+        
+        echo "All commands are available."
     fi
 }
 
--- a/.scripts/package_manager_run.sh
+++ b/.scripts/package_manager_run.sh
@@ -29,8 +29,6 @@ package_manager_run() {
         if [[ "$(docker compose 2>&1)" == *"docker: 'compose' is not a docker command."* ]]; then
             fatal "The 'docker compose' command is not functional. Follow the directions at https://docs.docker.com/compose/install/linux/ to install compose."
         fi
-
-
     fi
 }
 
--- a/.scripts/package_manager_run.sh
+++ b/.scripts/package_manager_run.sh
@@ -17,7 +17,7 @@ package_manager_run() {
         echo "Supported package manager not detected. Checking for dependencies ..."
         # Define an array of commands
         commands=("curl" "docker" "docker-compose" "git" "grep" "sed" "whiptail")
-        
+
         # Iterate over each command in the array
         for cmd in "${commands[@]}"; do
             # Check if the command is available in the system
@@ -25,7 +25,7 @@ package_manager_run() {
                 fatal "Error: '$cmd' is not available. Exiting..."
             fi
         done
-        
+
         echo "All commands are available."
     fi
 }
--- a/.scripts/package_manager_run.sh
+++ b/.scripts/package_manager_run.sh
@@ -12,11 +12,11 @@ package_manager_run() {
         run_script "pm_pacman_${ACTION}"
     elif [[ -n "$(command -v yum)" ]]; then
         run_script "pm_yum_${ACTION}"
-    else
+    elif [[ "${ACTION}" == "install" ]]; then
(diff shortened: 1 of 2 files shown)


.scripts/package_manager_run.sh
```diff
@@ -13,7 +13,20 @@ package_manager_run() {
     elif [[ -n "$(command -v yum)" ]]; then
         run_script "pm_yum_${ACTION}"
     else
-        fatal "Supported package manager not detected!"
+        # We might not need a supported package manager at all if the dependencies are there already. Let's validate that.
```
nemchik, line 16: This whole section (all the changes in this file) might not work as this script is called multiple times sequentially by other scripts, such as here:
https://github.com/GhostWriters/DockSTARTer/blob/55fdbccd2d1549d44996115d7aca5971fe901ce6/.scripts/update_system.sh#L6-L9

It might work if we nest an additional `if` inside this `else` to check the `${ACTION}` and only act when it's called with `install` (most of the deps you have below) or `install_docker` (special handling to check for `docker` and the `docker compose` v2 plugin).

.scripts/symlink_ds.sh
```diff
@@ -5,25 +5,40 @@ IFS=$'\n\t'
...
-    # /usr/bin/ds
-    if [[ -L "/usr/bin/ds" ]] && [[ ${SCRIPTNAME} != "$(readlink -f /usr/bin/ds)" ]]; then
-        info "Attempting to remove /usr/bin/ds symlink."
-        sudo rm -f "/usr/bin/ds" || fatal "Failed to remove file.\nFailing command: ${F[C]}sudo rm -f \"/usr/bin/ds\""
-    fi
-    if [[ ! -L "/usr/bin/ds" ]]; then
-        info "Creating /usr/bin/ds symbolic link for DockSTARTer."
-        sudo ln -s -T "${SCRIPTNAME}" /usr/bin/ds || fatal "Failed to create symlink.\nFailing command: ${F[C]}sudo ln -s -T \"${SCRIPTNAME}\" /usr/bin/ds"
-    fi
+    if findmnt -n /usr | grep "ro"; then
```
nemchik, line 8: This is a clever approach, but I wonder if we can set a variable along the lines of `DS_SYMLINK_TARGET` with a default value of `/usr/local/bin/ds` and then change the value to `$HOME/bin/ds` if `/usr` is not writeable, and then let the rest of the logic be the same for both cases. There's also handling for `/usr/bin/ds` in the existing code, and I believe the original need for handling `/usr/bin/ds` vs `/usr/local/bin/ds` is which one is on path for different OS, so it could even be worth checking the path variable to see which one we need, or if we have to warn the user to adjust their path.

nemchik (changes requested):

.scripts/package_manager_run.sh
```diff
@@ -13,7 +13,20 @@ package_manager_run() {
...
+        echo "Supported package manager not detected. Checking for dependencies ..."
+        # Define an array of commands
+        commands=("curl" "docker" "docker-compose" "git" "grep" "sed" "whiptail")
```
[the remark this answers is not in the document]
dylanmtaylor: I'll remove the check for it then.

.scripts/package_manager_run.sh
dylanmtaylor: I have tested this and in practice, it works and there is no noticeable latency with other commands, but in the interest of it being fully correct I can add that condition

.scripts/symlink_ds.sh
dylanmtaylor: Hmm, I suppose we can do something with that path variable. This is what mine looks like as an example.

$ echo $PATH
/var/home/dylan/.local/bin:/var/home/dylan/bin:/usr/local/bin:/usr/bin:/usr/local/sbin:/usr/sbin

Notably, all of the possible paths are on it. I'll try to do a check to decide if we want to print that message.

.scripts/package_manager_run.sh
```diff
@@ -13,7 +13,20 @@ package_manager_run() {
...
+
+        # Iterate over each command in the array
+        for cmd in "${commands[@]}"; do
+            # Check if the command is available in the system
+            if ! command -v "$cmd" &> /dev/null; then
+                fatal "Error: '$cmd' is not available. Exiting..."
+            fi
+        done
+
+        echo "All commands are available."
```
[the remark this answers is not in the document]
dylanmtaylor: I removed these lines as it felt unnecessarily verbose.

.scripts/package_manager_run.sh
```diff
@@ -12,8 +12,23 @@ package_manager_run() {
         run_script "pm_pacman_${ACTION}"
     elif [[ -n "$(command -v yum)" ]]; then
         run_script "pm_yum_${ACTION}"
-    else
-        fatal "Supported package manager not detected!"
+    elif [[ ${ACTION} == "install" ]]; then
```
nemchik, line 15: This also needs to handle `install_docker` as an `ACTION` which is specific to ensuring `docker` and `docker compose` are available.

.scripts/package_manager_run.sh
```diff
@@ -12,8 +12,23 @@ package_manager_run() {
...
+        # We might not need a supported package manager at all if the dependencies are there already. Let's validate that.
+
+        # Define an array of commands
+        commands=("curl" "docker" "git" "grep" "sed" "whiptail")
```
nemchik, line 19: Check for docker separately with the `install_docker` action.

.scripts/package_manager_run.sh
```diff
@@ -12,8 +12,23 @@ package_manager_run() {
...
+
+        # Iterate over each command in the array
+        for cmd in "${commands[@]}"; do
+            # Check if the command is available in the system
+            if ! command -v "$cmd" &> /dev/null; then
+                fatal "Error: '$cmd' is not available. Exiting..."
+            fi
+        done
+
+        if [[ "$(docker compose 2>&1)" == *"docker: 'compose' is not a docker command."* ]]; then
```
nemchik, line 29: would this work with `if ! docker compose version > /dev/null 2>&1; then` ?

.scripts/symlink_ds.sh
```diff
@@ -5,25 +5,26 @@ IFS=$'\n\t'
...
-    # /usr/local/bin/ds
-    if [[ -L "/usr/local/bin/ds" ]] && [[ ${SCRIPTNAME} != "$(readlink -f /usr/local/bin/ds)" ]]; then
-        info "Attempting to remove /usr/local/bin/ds symlink."
-        sudo rm -f "/usr/local/bin/ds" || fatal "Failed to remove file.\nFailing command: ${F[C]}sudo rm -f \"/usr/local/bin/ds\""
-    fi
-    if [[ ! -L "/usr/local/bin/ds" ]]; then
-        info "Creating /usr/local/bin/ds symbolic link for DockSTARTer."
-        sudo ln -s -T "${SCRIPTNAME}" /usr/local/bin/ds || fatal "Failed to create symlink.\nFailing command: ${F[C]}sudo ln -s -T \"${SCRIPTNAME}\" /usr/local/bin/ds"
-    fi
+    for target in "${ds_symlink_targets[@]}"; do
```
nemchik, line 18: This is nearly perfect! But let's stick to uppercase variable naming, and single/plural similarities. ex: `DS_SYMLINK_TARGETS` (for the array) and `DS_SYMLINK_TARGET` (for the iteration inside the loop) just in case there's some other `target` variable in the environment that could conflict.

.scripts/symlink_ds.sh
```diff
@@ -5,25 +5,26 @@ IFS=$'\n\t'
...
+        if [[ -L ${target} ]] && [[ ${SCRIPTNAME} != "$(readlink -f "${target}")" ]]; then
+            info "Attempting to remove ${target} symlink."
+            sudo rm -f "${target}" || fatal "Failed to remove file. Failing command: sudo rm -f \"${target}\""
```
nemchik, line 21:
```suggestion
            sudo rm -f "${target}" || fatal "Failed to remove file.\nFailing command: ${F[C]}sudo rm -f \"${target}\""
```
Keep the newline and color variable

.scripts/symlink_ds.sh
```diff
@@ -5,25 +5,26 @@ IFS=$'\n\t'
...
+        fi
+        if [[ ! -L ${target} ]]; then
+            info "Creating ${target} symbolic link for DockSTARTer."
+            sudo ln -s -T "${SCRIPTNAME}" "${target}" || fatal "Failed to create symlink. Failing command: sudo ln -s -T \"${SCRIPTNAME}\" \"${target}\""
```
nemchik, line 25:
```suggestion
            sudo ln -s -T "${SCRIPTNAME}" "${target}" || fatal "Failed to create symlink.\nFailing command: ${F[C]}sudo ln -s -T \"${SCRIPTNAME}\" \"${target}\""
```
Keep the newline and color variable

.scripts/symlink_ds.sh
```diff
@@ -5,25 +5,26 @@ IFS=$'\n\t'
...
-    fi
-    if [[ ! -L "/usr/bin/ds" ]]; then
-        info "Creating /usr/bin/ds symbolic link for DockSTARTer."
-        sudo ln -s -T "${SCRIPTNAME}" /usr/bin/ds || fatal "Failed to create symlink.\nFailing command: ${F[C]}sudo ln -s -T \"${SCRIPTNAME}\" /usr/bin/ds"
+    if findmnt -n /usr | grep "ro" > /dev/null; then
+        if [[ $PATH != *"$HOME/bin"* ]]; then
+            warn "Read only /usr filesystem detected. Symlinks will be created in $HOME/bin. You will need to add this to your path."
+        fi
+        mkdir -p "$HOME/bin" # Make sure the path exists.
+        ds_symlink_targets=("HOME/bin")
```
nemchik, line 13: Does this need a `$` ?

nemchik (changes requested):

.scripts/symlink_ds.sh
dylanmtaylor: Yes, great catch!

.scripts/package_manager_run.sh
dylanmtaylor: Maybe. I think the question is dependent on the return value being 0 or 1.

dylanmtaylor: I'm not sure, so don't want to change this right now.

dylanmtaylor:
```
dylan@Dylan-PowerEdge-R730:~$ docker compose version
docker: 'compose' is not a docker command.
See 'docker --help'
dylan@Dylan-PowerEdge-R730:~$ echo $?
1
```

Decided to try this after all. Doesn't break things that are already running to mess up compose.

dylanmtaylor: $ if ! docker compose version > /dev/null 2>&1; then echo "fail"; fi
fail

I think we can do it that way.

dylanmtaylor: Yes, this should work.
```
dylan@Dylan-PowerEdge-R730:~$ docker compose version
Docker Compose version v2.24.5
dylan@Dylan-PowerEdge-R730:~$ echo $?
0
dylan@Dylan-PowerEdge-R730:~$ if ! docker compose version > /dev/null 2>&1; then echo "fail"; fi
```

nemchik:
Changing branch and merging so I can stage in some additional adjustments.

PR #1732, opened 2024-02-23
OS support

nemchik:
# Pull request

**Purpose**
Follow up from https://github.com/GhostWriters/DockSTARTer/pull/1731 to make DS compatible with CoreOS (or any OS where a supported package manager is not detected.

**Approach**
This results in DS exiting with a `fatal` when any required dependency is missing (and a supported package manager is not found to install the dependency) and informing the user that they need to install said dependency manually.

**Requirements**


dylanmtaylor (approved):

dylanmtaylor:
The only thing I'm not a fan of is no longer providing the user the link for how to install compose.

nemchik:
> The only thing I'm not a fan of is no longer providing the user the link for how to install compose.

Yeah I thought a bit about removing that and decided it would be nice to keep it the same as the others in terms of wording and verbosity. Technically there are many ways one could install compose as well as any of the other dependencies.

dylanmtaylor:
That page gives a bunch of different options. Including the exact command you need on coreos.

dylanmtaylor:
If you use two targets for home, if either isn't in the path, it looks like it will warn even if nothing is wrong and it would work fine.

dylanmtaylor:
Can we add the docker compose doc I linked to the readme at least? It's probably exactly what the user needs to do.

nemchik:
> If you use two targets for home, if either isn't in the path, it looks like it will warn even if nothing is wrong and it would work fine.

This is true. I also found many OS don't include `$HOME/bin` or `$HOME/.local/bin` in `$PATH` by default unless it exists before when their bash profile is loaded, so in theory some users might just need to log out and back in or reboot and there would not be an issue. I am open to suggestions on the wording.

> Can we add the docker compose doc I linked to the readme at least? It's probably exactly what the user needs to do.

I am definitely fine with the link being added to the readme with a description somewhere, and I'm not dead set that it can't be included in the failing command output, but maybe it would be cleaner as a `warn` right before the `fatal`, what do you think?

maybe
```shell
            if ! docker compose version > /dev/null 2>&1; then
                warn "Please see https://docs.docker.com/compose/install/linux/ to install 'docker compose'"
                fatal "Error: 'docker compose' is not available. Please install 'docker compose' and try again."
            fi
```

PR #1739, opened 2024-03-05
Fix some incorrect and missing variables

CLHatch:
Fix some incorrect and missing variables

# Pull request

**Purpose**

**Approach**

**Learning**
Describe the research stage

**Requirements**


compose/.apps/plex/plex.yml
```diff
@@ -3,7 +3,7 @@ services:
     container_name: plex
     environment:
       - PGID=${PGID}
-      - PLEX_CLAIM=${PLEX_ENVIRONMENT_CLAIM}
+      - PLEX_CLAIM=${PLEX_ENVIRONMENT_PLEX_CLAIM}
```
nemchik, line 6: This isn't really a problem the way it was. As long as the label and variable match (which they did) and the left side of the `=` in this file is correct then it's fine.

nemchik (commented):
Added a comment on the plex var.

I mentioned in discord that I have some concerns with the number of vars that gluetun seems to have. I don't know if all of them are required, but often with other apps we've tried to make everything available, or at least the most common use cases. I'm not too worried about the `.env` file size, but it can make things a bit confusing as to which vars you need to bother with. I'm a bit more concerned with the terminal menu and how (or even if) whiptail will handle this many vars. Also the user experience of opting in to editing the vars will be a long process, and has the same issue of not being sure which vars to bother with.

I'm not saying this can't land, but want feedback on whether it's feasible to trim down or if we need them all.

compose/.apps/plex/plex.yml
CLHatch: Right, knew this one wasn't really a "problem".  I just renamed it to be consistent with the new `<APPNAME>_ENVIRONMENT_<VARIABLE>` naming scheme.

CLHatch:
> Added a comment on the plex var.
> 
> I mentioned in discord that I have some concerns with the number of vars that gluetun seems to have. I don't know if all of them are required, but often with other apps we've tried to make everything available, or at least the most common use cases. I'm not too worried about the `.env` file size, but it can make things a bit confusing as to which vars you need to bother with. I'm a bit more concerned with the terminal menu and how (or even if) whiptail will handle this many vars. Also the user experience of opting in to editing the vars will be a long process, and has the same issue of not being sure which vars to bother with.
> 
> I'm not saying this can't land, but want feedback on whether it's feasible to trim down or if we need them all.

Not all of them are used all the time.  Here is where they are all described:
https://github.com/qdm12/gluetun-wiki/tree/main/setup/options
https://github.com/qdm12/gluetun-wiki/tree/main/setup/providers

To whittle them down a bit, I could see most people not needing the `DNS` ones at all, so could likely drop those, and anyone that does need them could do an override:
https://github.com/qdm12/gluetun-wiki/blob/main/setup/options/dns.md

Possibly the same for the healthcheck variables:
https://github.com/qdm12/gluetun-wiki/blob/main/setup/options/healthcheck.md

I could also see not many people needing to be able to change the three internal listen ports, so could likely drop these as well, which are `HTTP_CONTROL_SERVER_ADDRESS`, `HTTPPROXY_LISTENING_ADDRESS`, and `SHADOWSOCKS_LISTENING_ADDRESS`.  Most would likely keep those at the defaults, which are the three ports we expose.

PR #1816, opened 2024-12-08
Fix filebrowser healthcheck

vsilvar:
# Pull request

**Purpose**
File browser healthchecks were always failing due to using the incorrect port.
This was caused by a change where the `FB_PORT` env variable is no longer set.

**Approach**
As the FB_PORT env variable is no longer set, internally, within the container, port 80 is now always used.
So, as health checks run inside the container, the internal port needs to be used and not the host port.

**Requirements**


nemchik (approved):
lgtm

PR #1854, opened 2025-03-23
Whisparr

SpiderManAPV:
# Pull request

**Purpose**
Adds the Whisparr app.

**Approach**
It adds the app

- [x] Verify the app launches successfully.

**Learning**
N/A

**Requirements**

### DIFF
--- a/compose/.apps/whisparr/whisparr.aarch64.yml
+++ b/compose/.apps/whisparr/whisparr.aarch64.yml
@@ -1,0 +1,3 @@
+services:
+  whisparr:
+    image: hotio/whisparr:${WHISPARR_TAG}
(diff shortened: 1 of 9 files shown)


angrycuban13:
Whisparr in its current state isn't very usable and is seeing very little development.

SpiderManAPV:
That's fine if it was excluded for a reason. Just figured I'd try to help get all the main *arrs added.

nemchik:
I thought on this for a while and have decided to close the PR without merging.

The specific reason is:
For the most part DS has apps that are easy enough for anyone to run, which includes our support team in cases where a user needs help figuring something out about how DS or docker can be have a learning curve compared to running without docker. Even if nobody ever asks for help with this app, I don't want to open the door to the potential that our support team would be asked to support such an app (usually by running it themselves to confirm an issue or how things work). Granted they can simply not respond to such a request, but I'd rather not put anyone in an uncomfortable position while they are offering their free time to help others.

All that said, DS has always allowed users to put anything they want in their override file (which can include the entire compose for an app DS does not include), so anyone can run this app or any other app we don't include.

